PR #37, opened 2024-10-10
Add text asset support

kannwism:


Makuh17:
> We should make sure that one can create validation rapids

Was this addressed? I do not really know how to tell from the changes

PR #53, opened 2024-10-15
added the assets to the normal order creation. adjusted the quickstar…

LinoGiger:
…t & fixed some minor bugs

### DIFF
--- a/examples/transcription_order.py
+++ b/examples/transcription_order.py
@@ -21,17 +21,8 @@ def new_transcription_order(rapi: RapidataClient):
         ).add_transcription_rapid(
             asset=MediaAsset("examples/data/waiting.mp4"),
             question="Click any words that are part of a name",
-            transcription=[
-                "This",
-                "is",
-                "Mr.",
-                "Bean",
-                "waiting",
-                "for",
-                "his",
-                "friends",
-            ],
-            correct_words=["Mr.", "Bean"],
+            transcription="This is Mr. Bean waiting for his friends",
+            truths=[2, 3],
         )
     ).create()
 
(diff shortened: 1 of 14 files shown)


examples/transcription_order.py
```diff
@@ -21,17 +21,8 @@ def new_transcription_order(rapi: RapidataClient):
...
-                "Mr.",
-                "Bean",
-                "waiting",
-                "for",
-                "his",
-                "friends",
-            ],
-            correct_words=["Mr.", "Bean"],
+            transcription="This is Mr. Bean waiting for his friends",
+            truths=[2, 3],
```
Makuh17, line 25: nice catch, I completely missed this with my change

src/rapidata/rapidata_client/dataset/rapidata_dataset.py
```diff
@@ -22,7 +23,8 @@ def __init__(self, dataset_id: str, openapi_service: OpenAPIService):
         self.openapi_service = openapi_service
         self.local_file_service = LocalFileService()
 
-    def add_texts(self, texts: list[str]):
+    def add_texts(self, text_assets: list[TextAsset]):
```
Makuh17, line 26: ah, i guess here it would be an issue if we wanted a multiasset with text

src/rapidata/rapidata_client/order/rapidata_order_builder.py
```diff
@@ -129,8 +131,12 @@ def create(self, submit: bool = True, max_workers: int = 10) -> RapidataOrder:
         if isinstance(
             self._workflow, CompareWorkflow
         ):  # Temporary fix; will be handled by backend in the future
+            assert all(isinstance(item, MultiAsset) for item in self._assets), (
+                "The media paths must be of type MultiAsset for comparison tasks."
+            )
+            media_paths = cast(list[MultiAsset], self._assets)
```
Makuh17, line 137: This assumes that the multiassets contain only media assets right?

Makuh17 (approved):
Generally looks good i would say

src/rapidata/rapidata_client/order/rapidata_order_builder.py
LinoGiger: yes, atm this is all we need. but again we could just change the checks for that. it would not be a breaking change

PR #67, opened 2024-10-17
add more filters

Makuh17:


src/rapidata/rapidata_client/filter/campaign_filter.py
```diff
@@ -0,0 +1,17 @@
...
+    CampaignUserFilterModel,
+)
+
+
+class CampaignFilter(Filter):
+
+    def __init__(self, campaign_ids: list[str]):
+        self.campaign_ids = campaign_ids
+
+    def to_model(self) -> Any:
```
kannwism, line 13: Not sure if it's necessary to provide the `Any` here. It might overwrite the return type to Any, even if it knows what kind of model it is returning. Same for the other child classes

kannwism (approved):

src/rapidata/rapidata_client/filter/age_filter.py
```diff
@@ -0,0 +1,23 @@
+from typing import Any
+from rapidata.rapidata_client.filter.base_filter import Filter
+from rapidata.api_client.models.age_user_filter_model import AgeUserFilterModel
+from rapidata.api_client.models.age_group import AgeGroup
+
+
+class AgeFilter(Filter):
+
+    def __init__(self, age_groups: list[str]):
```
LinoGiger, line 9: why are we not directly providing a list of age groups here?

src/rapidata/rapidata_client/filter/campaign_filter.py
Makuh17: Ah okay. I just copied that from somewhere else

src/rapidata/rapidata_client/filter/gender_filter.py
```diff
@@ -0,0 +1,23 @@
+from typing import Any
+from rapidata.rapidata_client.filter.base_filter import Filter
+from rapidata.api_client.models.gender_user_filter_model import GenderUserFilterModel
+from rapidata.api_client.models.gender import Gender
+
+
+class GenderFilter(Filter):
+
+    def __init__(self, genders: list[str]):
+        # check that all the age groups exist in the AgeGroup enum
```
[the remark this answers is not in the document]
Makuh17: Ah yes, good catch

src/rapidata/rapidata_client/filter/age_filter.py
Makuh17: Age group is part of the generated api which I thought we generally did not want to expose. Not sure, maybe it makes sense to expose it?

src/rapidata/rapidata_client/filter/user_score_filter.py
```diff
@@ -0,0 +1,19 @@
+from typing import Any
+from rapidata.rapidata_client.filter.base_filter import Filter
+from rapidata.api_client.models.user_score_user_filter_model import (
+    UserScoreUserFilterModel,
+)
+
+
+class UserScoreFilter(Filter):
+
+    def __init__(self, upper_bound: int, lower_bound: int):
```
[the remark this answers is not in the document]
Makuh17: Sure, that's a good idea

PR #85, opened 2024-10-23
Add evlaluation workflow support

kannwism:


tests/integration/test_example_orders.py
```diff
@@ -51,16 +51,22 @@ def test_quickstart(self):
         CLIENT_ID = os.getenv("CLIENT_ID")
         CLIENT_SECRET = os.getenv("CLIENT_SECRET")
         ENDPOINT = os.getenv("ENDPOINT")
+        CERT_PATH = os.getenv("CERT_PATH")
+        TOKEN_URL = os.getenv("TOKEN_URL")
```
Makuh17, line 55: What is this now?

tests/integration/test_internal_orders.py
```diff
@@ -0,0 +1,34 @@
...
+                truths=["Marsupial"],
+                metadata=[PromptMetadata(prompt="Hint: It has a pouch")],
+            )
+            .create()
+        )
+
+        order = (
+            self.rapi.new_order("Test Evaluation Order")
+            .workflow(EvaluationWorkflow(validation_set_id=validation_set.id))
+            .media([MediaAsset(path="examples/data/wallaby.jpg")])
```
Makuh17, line 29: Why do we need to add media here?

Makuh17 (commented):
I am a little unsure about why we add media when creating the order?

tests/integration/test_internal_orders.py
kannwism: Not necessary anymore

PR #87, opened 2024-10-28
added functionality to find orders by name

LinoGiger:


src/rapidata/rapidata_client/order/rapidata_order_builder.py
```diff
@@ -143,11 +143,12 @@ def create(self, submit: bool = True, max_workers: int = 10) -> RapidataOrder:
         )
 
         self.order_id = result.order_id
-        self._dataset = RapidataDataset(result.dataset_id, self._openapi_service)
+        self._dataset = RapidataDataset(result.dataset_id, self._openapi_service) # type: ignore - should only be None if order is evaluation order
```
kannwism, line 146: Would something like this be cleaner than a type ignore?

```suggestion
        self._dataset = RapidataDataset(result.dataset_id, self._openapi_service)  if result.dataset_id else None
```

src/rapidata/rapidata_client/rapidata_client.py
```diff
@@ -85,12 +91,38 @@ def get_order(self, order_id: str) -> RapidataOrder:
         # TODO: check the pipeline for the dataset id - not really necessary atm
         # order = self.openapi_service.order_api.order_get_by_id_get(order_id)
         # pipeline = self.openapi_service..pipeline_get_by_id_get(order.pipeline_id)
+        try:
+            order = self.openapi_service.order_api.order_get_by_id_get(order_id)
+        except Exception:
+            raise ValueError(f"Order with ID {order_id} not found.")
```
kannwism, line 97: Nit: Should we still print the original error message as well to make debugging easier in case this fails?

src/rapidata/rapidata_client/rapidata_client.py
```diff
@@ -85,12 +91,38 @@ def get_order(self, order_id: str) -> RapidataOrder:
...
+            amount (int, optional): The amount of orders to return. Defaults to 1.
+
+        Returns:
+            list[RapidataOrder]: A list of RapidataOrder instances.
+        """
+        try:
+            order_page_result = self.openapi_service.order_api.order_query_get(QueryOrdersModel(
+                page=PageInfo(index=1, size=amount),
+                filter=RootFilter(filters=[Filter(field="OrderName", operator="Contains", value=name)])))
+        except Exception:
```
kannwism, line 120: Same Nit: here

PR #91, opened 2024-10-30
added resience to get_results and display_progress_bar

LinoGiger:


src/rapidata/rapidata_client/order/rapidata_order.py
```diff
@@ -86,31 +86,38 @@ def _get_workflow_id(self):
...
     def get_results(self):
         """
-        Gets the results of the order.
+        Gets the results of the order. 
+        If the order is not completed, this method will block until the order is completed and then return the results.
 
         :return: The results of the order.
         :rtype: dict
         """
+        while self.get_status().state != "Completed":
```
kannwism, line 118: We only want to enter the while loop if we're in state processing

kannwism (changes requested):

kannwism (approved):

PR #98, opened 2024-11-01
remove the str enum for earlier python versions

LinoGiger:


kannwism (approved):
Alright, then we could put the requirement for the python version back to 3.10, I increased it to 3.11 because of the StrEnum that Manuel introduced

PR #114, opened 2024-11-05
added name to validation set and possibility to find them

LinoGiger:
i also matched the naming of the validation set to the naming in the builder


kannwism:
Soo, in order to get a little bit used to not breaking stuff we should try making this backwards compatible, while emitting a warning.

I'd also be curious on how companies decided when to remove deprecated stuff in the end? Do they track usage somehow?

PR #115, opened 2024-11-07
Add daemon to refresh API token

Sneccello:
Refreshing API token periodically because the lifetime of the client can be more than the token expiration time (currently 30min)

Getting an API token in the constructor, because without it, technically it can fail if the first api call is before the thread spawn

jira: https://rapidata.atlassian.net/browse/RAPID-3280

### DIFF
--- a/src/rapidata/service/openapi_service.py
+++ b/src/rapidata/service/openapi_service.py
@@ -1,5 +1,7 @@
 import json
+import time
 import requests
+import threading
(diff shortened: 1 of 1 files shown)


LucStr (approved):

LinoGiger (approved):

LucStr:
I think having a full thread just asleep all the time is probably not the best solution, but I think it more than suffices for now. Thanks Marcell!

PR #131, opened 2024-11-14
feat: allow browser login

jorgeparavicini:


src/rapidata/rapidata_client/rapidata_client.py
```diff
@@ -27,10 +27,10 @@ class RapidataClient:
 
     def __init__(
         self,
-        client_id: str,
-        client_secret: str,
+        client_id: str | None,
+        client_secret: str | None,
         endpoint: str = "https://api.rapidata.ai",
-        token_url: str = "https://auth.rapidata.ai/connect/token",
```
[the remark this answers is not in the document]
LinoGiger: i'm also getting some errors here:
image

src/rapidata/service/token_manager.py
```diff
@@ -0,0 +1,137 @@
...
+    def auth_header(self):
+        return f"{self.token_type} {self.access_token}"
+
+    @property
+    def time_until_refresh(self) -> float:
+        return (self.expires_at - datetime.now()).total_seconds()
+
+    @property
+    def is_expired(self) -> bool:
+        return datetime.now() > self.expires_at
```
LinoGiger, line 28: in the loop below and in the loop here i would put in some buffer that it rather requests it too early than too late. it can happen now that we get an error if we request something right when it expires if i understand this correctly? maybe have a 1 minute buffer?

jorgeparavicini: It refetches the token once the existing token is expired at 80%, this method is more used as a last resort

src/rapidata/rapidata_client/rapidata_client.py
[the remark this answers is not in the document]
jorgeparavicini: > will this break existing sdk instances? like Uthana?

It shouldn't as I left the name the same and they should not be setting the token endpoint explicitly

src/rapidata/service/token_manager.py
LinoGiger: okay, but that thing is still wrapped in a min(30)?  here:
                        sleep_time = min(30, int(self._current_token.time_until_refresh * (1 - self._refresh_threshold)))

or does it get decided elsewhere?

jorgeparavicini: I completely rewrote the logic

LinoGiger (approved):

src/rapidata/service/credential_manager.py
```diff
@@ -0,0 +1,227 @@
...
+
+    def _write_credentials(
+        self, credentials: Dict[str, List[ClientCredential]]
+    ) -> None:
+        data = {
+            env: [cred.to_dict() for cred in creds]
+            for env, creds in credentials.items()
+        }
+
+        with open(self.config_path, "w") as f:  # type: TextIOWrapper
```
kannwism, line 94: Pylance gives this error here:
Type annotation not supported for this statementPylancereportInvalidTypeForm

But it doesn't complain if it's simply removed. Is PyCharm complaining here otherwise?

src/rapidata/service/credential_manager.py
```diff
@@ -0,0 +1,227 @@
...
+            return None
+
+        auth_url = f"{self.endpoint}/connect/authorize/external?clientId=rapidata-cli&scope=openid profile email&writeKey={bridge_endpoint.write_key}"
+        webbrowser.open(auth_url)
+
+        access_token = self._poll_read_key(bridge_endpoint.read_key)
+        if not access_token:
+            return None
+
+        client_id, client_secret, display_name = self._create_client(access_token)
```
kannwism, line 215: If `_create_client()` returns None this will throw an error, as None can't be unpacked into the three values

kannwism (changes requested):

LinoGiger:
image
shouldn't this thing work with creating a login page?

jorgeparavicini:
@LinoGiger the backend hasnt deployed yet

kannwism (approved):
Looks good now and already worked for me 😎  Is it because I already have a client?

jorgeparavicini:
> Looks good now and already worked for me 😎 Is it because I already have a client?

No it was missing the client_id for getting the initial access token, I had fixed that ^^

PR #144, opened 2024-11-21
Feat/rapid 3411 improve simple builders

LinoGiger:


kannwism (commented):
Big one

Not sure if the time effort stuff is water tight, at the moment it looks a little error prone.

Also, probably there is a way to reduce boilerplate in all the builders

PR #153, opened 2024-11-29
feat: print auth url if browser did not open

jorgeparavicini:


kannwism (changes requested):
Probably needs a `poetry add colorama` to include the dependency in the package. Other wise looks good 👍  @LinoGiger Maybe we can use the coloring in other places as well down the road

jorgeparavicini:
i would mainly use the logger for logging stuff. This is a special case as I explicitly want the user to do something and per default the logging is disabled. Mandatory: python sucks

kannwism (approved):
👍

PR #155, opened 2024-12-02
Fix evaluation workflow typing issue

kannwism:


LinoGiger:
where did that change go?

LinoGiger (approved):

kannwism:
Yeah I'm not 100% sure what's going on, but it's not on main. Was it accidentally reverted when we experimented with pyright?

PR #156, opened 2024-12-03
Add custom X-Client header to calls

kannwism:


src/rapidata/service/openapi_service.py
```diff
@@ -74,3 +80,16 @@ def workflow_api(self) -> WorkflowApi:
...
+        Returns the version of the currently installed rapidata package.
+
+        Returns:
+            Optional[str]: The version of the installed rapidata package,
+                        or None if the package is not found
+        """
+        try:
+            return version("rapidata")
+        except PackageNotFoundError:
+            return None
```
LinoGiger, line 95: this gives me a PackageNotFoundError? am i running it wrong?

kannwism: Huh, that makes no sense to me. Could you try running `rapi.openapi_service._get_rapidata_package_version()` ?

It makes no sense, since the error can't really occur. The function is in the rapidata package, so if it's being executed it kinda needs to be installed.

And even then, this should actually catch the error and return None in that case

LinoGiger: no it's a handeled error, but that means that the package version is just null

LinoGiger: i'm assuming that the error is due to the fact that i'm connecting to my local rapidata version though a kernel where i do not have rapidata installed. however when i have it installed it does not have the method yet. idk how you tested it or how i can test it.

also - is it intended that, running it locally will just be null instead of some default or the current version number?

kannwism: If I go on main, do `poetry install` and run a test, there is no x-client thing being sent. If I go to this branch, do `poetry install` and run a test it sends `'X-Client' = ''RapidataPythonSDK/1.10.1'`. So to me it seems that it is working as intended?

PR #158, opened 2024-12-13
Feat/rapid 3591 support locate and line

LinoGiger:


src/rapidata/rapidata_client/settings/free_text_minimum_characters.py
```diff
@@ -0,0 +1,16 @@
...
+class FreeTextMinimumCharacters(RapidataSetting):
+    """
+    Set the minimum number of characters a user has to type.
+    
+    Args:
+        value (int): The minimum number of characters for free text.
+    """
+
+    def __init__(self, value: int):
+        if value < 0:
```
Makuh17, line 12: would argue this should be 1

src/rapidata/rapidata_client/settings/free_text_minimum_characters.py
```diff
@@ -0,0 +1,16 @@
...
+            raise ValueError("The minimum number of characters must be greater than or equal to 0.")
+        if value > 20:
```
Makuh17, line 14: should this be slightly higher?

src/rapidata/rapidata_client/validation/_validation_set_builder.py
```diff
@@ -127,6 +139,9 @@ def __add_classify_rapid(
         Raises:
             ValueError: If the lengths of categories and truths are inconsistent.
         """
+        # if not all(truth in answer_options for truth in truths):
+        #     raise ValueError("Truths must be part of the answer options")
+
```
Makuh17, line 144: why is this commented?

Makuh17 (approved):
Went through it rather quickly, but based on my usage, it seems to work as intended

PR #179, opened 2025-01-23
adjusted the select words validation

LinoGiger:


src/rapidata/rapidata_client/validation/rapids/rapids_manager.py
```diff
@@ -91,7 +91,9 @@ def select_words_rapid(self,
             truths: list[int],
             datapoint: str,
             sentence: str,
-            strict_grading: bool = True,
+            required_precision: float,
```
LucStr, line 94: I think we should add some sane default values here as people would just be confused

LucStr: just like down below

LucStr (approved):

PR #186, opened 2025-02-10
added ability to add new rapids to validation set

LinoGiger:


src/rapidata/rapidata_client/validation/rapids/rapids_manager.py
```diff
@@ -132,7 +132,7 @@ def select_words_rapid(self,
         asset = MediaAsset(datapoint)
         transcription_words = [
             TranscriptionWord(word=word, wordIndex=i)
-            for i, word in enumerate(sentence)
+            for i, word in enumerate(sentence.split(" "))
```
kannwism, line 135: Was that a bug that got caught?

kannwism (approved):
looks good

src/rapidata/rapidata_client/validation/rapids/rapids_manager.py
LinoGiger: yes exactly

PR #187, opened 2025-02-10
Feat/add ab test selection

LucStr:


src/rapidata/rapidata_client/order/_rapidata_order_builder.py
```diff
@@ -92,7 +92,7 @@ def _to_model(self) -> CreateOrderModel:
                 else None
             ),
             selections=[
-                CappedSelectionSelectionsInner(selection._to_model())
+                AbTestSelectionAInner(selection._to_model())
```
LinoGiger, line 95: why is this now the default?

LinoGiger (changes requested):
so everything is now an ab test per default but only the A part gets filled out unless you supply an ab test selection? but the n if you want an AB test you supply an AB test selection inside an AB test selection? - i don't really understand how that makes sense?

LinoGiger (approved):

PR #192, opened 2025-02-12
Feat/rapid 3768/ranking order

Sneccello:
Added API to support creating ranking orders


examples/ranking_order.py
```diff
@@ -0,0 +1,39 @@
+"""
+Ranking order
+"""
+from rapidata.rapidata_client import RapidataClient
+
+
+def new_ranking_order(rapi: RapidataClient):
+    rabbits = [
```
LinoGiger, line 8: adjust the name here to be more intuitive. like "datapoints" or "datapoint_links"

src/rapidata/rapidata_client/order/rapidata_order_manager.py
```diff
@@ -375,6 +374,65 @@ def create_locate_order(self,
...
+    def create_ranking_order(self,
+                             name: str,
+                             instruction: str,
+                             datapoints: list[str],
+                             total_comparison_budget: int,
+                             random_comparisons_ratio: float = 0.5,
+                             elo_start: int = 1200,
+                             elo_k_factor: int = 40,
+                             elo_scaling_factor: int = 400,
+                             responses_per_comparison: int = 10,
```
LinoGiger, line 386: i feel like this one is more important the random comparison ratio / elo start / k factor and elo scaling? - can you move it up if you agree? (also adjust the description then)

src/rapidata/rapidata_client/order/rapidata_order_manager.py
```diff
@@ -375,6 +374,65 @@ def create_locate_order(self,
             settings=settings
         )
 
+    def create_ranking_order(self,
```
LinoGiger, line 377: i don't understand why you decided to put this inbetween the locate & the draw. please either put it above both or below both

src/rapidata/rapidata_client/workflow/_ranking_workflow.py
```diff
@@ -0,0 +1,42 @@
...
+                 criteria: str,
+                 total_comparison_budget: int,
+                 random_comparisons_ratio,
+                 starting_elo: int,
+                 k_factor: int,
+                 scaling_factor: int,
+                 ):
+        super().__init__(type="CompareWorkflowConfig")
+        assert total_comparison_budget > 0, 'The comparison budget must be positive'
+        assert 0 <= random_comparisons_ratio <= 1, 'The ratio of randomly arranged matches should be between 0 and 1 (inclusive)'
```
LinoGiger, line 18: why are there no assertions for the rest? can k_factor be negative? starting elo negative? etc?

examples/ranking_order.py
LinoGiger: also maybe add it as an example for the frontpage?

Sneccello: > also maybe add it as an example for the frontpage?

where is that

LinoGiger: in the docs directory you can see the index.md file. you can there see how the examples are usually built up

LinoGiger (approved):
you can add it to the examples if you want, otherwise lgtm

PR #195, opened 2025-02-17
created rapidata results to convert results to pandas - open to furth…

LinoGiger:
…er functions in the future.

### DIFF
--- a/src/rapidata/rapidata_client/order/rapidata_results.py
+++ b/src/rapidata/rapidata_client/order/rapidata_results.py
@@ -1,0 +1,94 @@
+import pandas as pd
+
+class RapidataResults(dict):
+    def to_pandas(self) -> pd.DataFrame:
+        """
+        Converts the results to a pandas DataFrame, optimized for consistent structure.
+        """
+        if "results" not in self or not self["results"]:
+            return pd.DataFrame()
+        
+        if self["info"].get("orderType") == "Compare":
+            return self._compare_to_pandas()
+        
+        if self["info"].get("orderType") is None:
+            print("Warning: Results are old and Order type is not specified. Dataframe might be wrong.")
+
+        # Get the structure from first item
+        first_item = self["results"][0]
+        columns = []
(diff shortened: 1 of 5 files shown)


LinoGiger:
i intentionally didn't add it to the readme yet, because i first want to test it a bit in production and see if stuff needs to be adjusted

src/rapidata/rapidata_client/order/rapidata_results.py
```diff
@@ -0,0 +1,143 @@
...
+            path: List of keys forming the path to the desired value
+            
+        Returns:
+            The value at the specified path, or None if the path doesn't exist
+        """
+        for key in path[:-1]:
+            d = d.get(key, {})
+        return d.get(path[-1])
+
+    def _compare_to_pandas(self):
```
kannwism, line 100: Nit: Not quite clear what parts of compare are unique and what overlaps with the general case. In theory it shouldn't be too different from a classify with 2 options in my mind. This seems like there is quite some overlap with the general case above. Could this lead to accidental format differences, since these two are handled by completely different functions?

src/rapidata/rapidata_client/validation/validation_set_manager.py
```diff
@@ -318,7 +318,7 @@ def create_timestamp_set(self,
             name (str): The name of the validation set. (will not be shown to the labeler)
             instruction (str): The instruction to show to the labeler.
             truths (list[list[tuple[int, int]]]): The truths for each datapoint defined as start and endpoint based on miliseconds. 
-            Outher list is for each datapoint, inner list is for each truth.\n
+                outher list is for each datapoint, inner list is for each truth.\n
```
kannwism, line 321: What's outher?

kannwism (approved):

src/rapidata/rapidata_client/validation/validation_set_manager.py
LinoGiger: this was some pydantic issue i ignored for a while and just wanted to adjust

PR #200, opened 2025-02-20
Feat/rapid 4356 improve upload resiliance

LinoGiger:


src/rapidata/rapidata_client/assets/_media_asset.py
```diff
@@ -233,13 +235,21 @@ def __get_media_bytes(self, url: str) -> bytes:
             
         Raises:
             ValueError: If media type is unsupported or content validation fails
-            requests.exceptions.RequestException: If download fails
+            requests.exceptions.RequestException: If download fails after all retries
         """
+        # Use existing session or create new one
```
Makuh17, line 240: Is this comment true? Seems like it just raises an error, not creating a new one?

src/rapidata/rapidata_client/assets/_media_asset.py
```diff
@@ -233,13 +235,21 @@ def __get_media_bytes(self, url: str) -> bytes:
...
         try:
-            response = requests.get(url, stream=False)
+            response = self._session.get(
+                url, 
+                stream=False,
+                timeout=(5, 30)  # (connect timeout, read timeout)
+            )
             response.raise_for_status()
         except requests.exceptions.RequestException as e:
-            self._logger.error(f"Failed to download media from {url}: {str(e)}")
```
Makuh17, line 242: Just curious, why did we not see this error before? Or is it not printed to the user?

src/rapidata/rapidata_client/order/_rapidata_dataset.py
```diff
@@ -15,15 +15,25 @@
...
+from logging import Logger
+from requests.adapters import HTTPAdapter, Retry
+import time
+import requests
 
 
+def chunk_list(lst: list, chunk_size: int) -> Generator:
+    """Split list into chunks to prevent resource exhaustion"""
+    for i in range(0, len(lst), chunk_size):
+        yield lst[i:i + chunk_size]
```
Makuh17, line 28: I assume this also works even if the length of the array is not a multiple of chunk size? Or does it throw and error/wrap around?

Makuh17 (commented):
I think it looks fine, but tbh I am not too on top of what happens here. I will let @LucStr have the final word here

src/rapidata/rapidata_client/order/_rapidata_dataset.py
LinoGiger: l = [1,2,3,4,5,6]
print(l[0:10])
>[1, 2, 3, 4, 5, 6]

so it works fine

src/rapidata/rapidata_client/assets/_media_asset.py
LinoGiger: good catch, left over from tests

src/rapidata/rapidata_client/assets/_media_asset.py
```diff
@@ -233,13 +235,21 @@ def __get_media_bytes(self, url: str) -> bytes:
             
         Raises:
             ValueError: If media type is unsupported or content validation fails
-            requests.exceptions.RequestException: If download fails
+            requests.exceptions.RequestException: If download fails after all retries
         """
+        # Use existing session or throw error if not set
+        if self._session is None:
+            raise RuntimeError("HTTP session not configured")
```
LucStr, line 242: why do you throw an error here instead of creating a new session?

src/rapidata/rapidata_client/order/_rapidata_dataset.py
```diff
@@ -69,60 +79,147 @@ def upload_text_datapoint(text_asset: TextAsset | MultiAsset, index: int) -> Non
...
-
+        adapter = HTTPAdapter(
+            pool_connections=max_workers * 2,
+            pool_maxsize=max_workers * 4,
+            max_retries=retries
+        )
+        session.mount('http://', adapter)
+        session.mount('https://', adapter)
+
+        def upload_datapoint(
```
LucStr, line 125: I find it a bit unintuitive to have the method within the method just placed in the middle, I think it would make sense to move it to the bottom, what do you think?

src/rapidata/rapidata_client/order/_rapidata_dataset.py
```diff
@@ -69,60 +79,147 @@ def upload_text_datapoint(text_asset: TextAsset | MultiAsset, index: int) -> Non
...
-            upload_response = self.openapi_service.dataset_api.dataset_create_datapoint_post(
-                model=model,
-                files=files
-            )
-            if upload_response.errors:
-                raise ValueError(f"Error uploading datapoint: {upload_response.errors}")
-
+        adapter = HTTPAdapter(
+            pool_connections=max_workers * 2,
+            pool_maxsize=max_workers * 4,
```
LucStr, line 119: how did you get to these multipliers?

LucStr (approved):

src/rapidata/rapidata_client/order/_rapidata_dataset.py
LinoGiger: freeballed - they are apparently to help with the thread allocation

LinoGiger: but it doesn't really matter that much because it is anyways capped by the os / the max workers are not really accessible to people

src/rapidata/rapidata_client/assets/_media_asset.py
LinoGiger: i either create it outside of the media asset or inside, not both - so i went with outside to have some hyperparameters - would you prefer it inside?

src/rapidata/rapidata_client/order/_rapidata_dataset.py
LinoGiger: idk i think it's in the correct place, above is all the set up and then below is where the method actually gets called

src/rapidata/rapidata_client/assets/_media_asset.py
LucStr: Mhmm, I am not that big of a fan of the design here, since we are mixing responsibilities, but let's get something working

PR #203, opened 2025-02-20
decrease timeout for token refetch

LucStr:


LinoGiger (approved):
yeah seems good, let's hope this helps

PR #209, opened 2025-03-05
Feat/rapid 4406 upload url backend

LinoGiger:


LucStr (approved):
I really want to push luma to use this, so I approve it. However in this specific case the AI generated code in my opinion is very hard to maintain and it took me a while to fully understand what is going on. 

what I would have done differently is to have two progress bars, where you first have a bar to upload the datapoints and then one that processes them. similar to the way docker does it when you pull an image. I think in this approach a lot of effort was put into converting these two progress bars into a single one, which I think has an argument to be had for the user experience. 

One thing that was done, which I disagree with is the threading events. I see that they were used to tell each other that either one of them should stop, but this could also just be a single one that can be triggered, also I am unsure if you want the full upload to stop, if a single upload failed in the background processing.

PR #210, opened 2025-03-05
Feat/rapid 4449 improve docs

LinoGiger:


src/rapidata/rapidata_client/filter/rapidata_filters.py
```diff
@@ -14,17 +14,25 @@ class RapidataFilters:
...
         language (LanguageFilter): The LanguageFilter instance.
         custom_filter (CustomFilter): The CustomFilter instance.
+
+    Example:
+        ```python
+        from rapidata import CountryFilter, LanguageFilter
+        filters=[CountryFilter(["US", "DE"]), LanguageFilter(["en"])]
+        ```
+
+        This will make it so the order is only shown to users from the US and Germany that have their phone set to English.
```
kannwism, line 35:
```suggestion
        This ensures the order is only shown to users in the US and Germany whose phones are set to English.
```

src/rapidata/rapidata_client/selection/rapidata_selections.py
```diff
@@ -11,21 +11,22 @@ class RapidataSelections:
...
+        capped (CappedSelection): The CappedSelection instance.
     
+    Example:
+        ```python
+        from rapidata import LabelingSelection, ValidationSelection
+        selections=[ValidationSelection("your-validation-set-id", 1), 
+                    LabelingSelection(2)]
+        ```
+
+        This will make it so the annotators will have to solve 1 validation task followed by two labeling tasks.
```
kannwism, line 28:
```suggestion
        This will require annotators to complete one validation task followed by two labeling tasks.
```

kannwism (approved):
Left two suggestion, but looks good overall

PR #218, opened 2025-03-12
fixed creation of demographic with link

LinoGiger:


Makuh17 (approved):
Guess this is copied from somewhere else, so didnt look too close

PR #225, opened 2025-03-18
added the dimension stuff

LinoGiger:


src/rapidata/rapidata_client/filter/user_score_filter.py
```diff
@@ -13,14 +13,15 @@ class UserScoreFilter(RapidataFilter):
...
+        dimension (str): The dimension of the userScore to be considerd for the filter. Default is "global".
 
     Example: 
         ```python
         UserScoreFilter(0.5, 0.9)
         ```
         This will only show the order to users that have a UserScore of >=0.5 and <=0.9
     """
-    def __init__(self, lower_bound: float = 0.0, upper_bound: float = 1.0):
+    def __init__(self, lower_bound: float = 0.0, upper_bound: float = 1.0, dimension: str = "global"):
```
LucStr, line 24: can you by default not provide any dimension at all? this way we can control the default value simply from the backend and don't have to worry about the evaluators implementation.

src/rapidata/rapidata_client/validation/validation_set_manager.py
```diff
@@ -399,22 +411,24 @@ def create_timestamp_set(self,
                 )
             )
 
-        return self._submit(name=name, rapids=rapids, print_confirmation=print_confirmation)
+        return self._submit(name=name, rapids=rapids, dimensions=dimensions, print_confirmation=print_confirmation)
     
     def create_mixed_set(self,
         name: str,
         rapids: list[Rapid],
+        dimensions: list[str] = ["global"],
```
LucStr, line 419: I think similar story to the one above, let's let the backend dictate what a default value is and only update it if we explicitly set dimensions. This way we could change the default to something else without any old SDK implementations interferring

LucStr (changes requested):

LucStr (approved):

PR #230, opened 2025-03-20
fixed the dimensions

LinoGiger:
if you now don't provide the dimensions the rapids will look like this - meaning they will propagate with "global", right? @LucStr


LucStr (approved):
this is correct, but @LinoGiger please don't share our internal datastructure in a public github repository :sweat_smile:

PR #245, opened 2025-04-25
moved sessions creation to singleton

LinoGiger:


src/rapidata/rapidata_client/validation/rapidata_validation_set.py
```diff
@@ -40,35 +40,35 @@ def update_dimensions(self, dimensions: list[str]):
...
+
+    #     adapter = HTTPAdapter(
+    #         pool_connections=max_workers * 2,
+    #         pool_maxsize=max_workers * 4,
+    #         max_retries=retries
+    #     )
+    #     session.mount('http://', adapter)
+    #     session.mount('https://', adapter)
+
+    #     return session
```
LucStr, line 71: should probably be removed

LucStr (approved):

src/rapidata/rapidata_client/validation/rapidata_validation_set.py
LinoGiger: yep, wups

PR #258, opened 2025-05-19
added multi modality examples for the quickstart

LinoGiger:
image

### DIFF
--- a/src/rapidata/rapidata_client/selection/static_selection.py
+++ b/src/rapidata/rapidata_client/selection/static_selection.py
@@ -8,7 +8,7 @@ class StaticSelection(RapidataSelection):
     Given a list of RapidIds, theses specific rapids will be shown in order for every session.
     
     Args:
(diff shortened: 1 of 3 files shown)


kannwism (approved):

kannwism:
The first thing a user sees is still images, right?

src/rapidata/rapidata_client/selection/static_selection.py
```diff
@@ -8,7 +8,7 @@ class StaticSelection(RapidataSelection):
     Given a list of RapidIds, theses specific rapids will be shown in order for every session.
     
     Args:
-        selections (list[str]): List of rapid ids to show.
```
kannwism, line 11: Don't get this change, is it intended?

LinoGiger: the parameter is called rapid_ids not selections. it was an old copy past error

PR #276, opened 2025-06-12
added better description for the different filters/selections/settings

LinoGiger:


src/rapidata/rapidata_client/selection/rapidata_selections.py
```diff
@@ -13,12 +13,12 @@ class RapidataSelections:
     All Tasks are called a "Session". A session can contain multiple tasks of different types.
```
jorgeparavicini, line 13: Saying "All Tasks are called a 'Session'" is confusing. My first instinct, then, is to ask why call it a task and not a session, which makes no sense.  Something along the lines of "The group of tasks that a user receives is called a session."

src/rapidata/rapidata_client/filter/rapidata_filters.py
```diff
@@ -18,13 +18,13 @@ class RapidataFilters:
...
     
     Attributes:
-        user_score (UserScoreFilter): The UserScoreFilter instance.
-        age (AgeFilter): The AgeFilter instance.
-        country (CountryFilter): The CountryFilter instance.
-        gender (GenderFilter): The GenderFilter instance.
-        language (LanguageFilter): The LanguageFilter instance.
-        not_filter (NotFilter): The NotFilter instance.
-        or_filter (OrFilter): The OrFilter instance.
+        user_score (UserScoreFilter): Filters the order based on the user's score.
```
jorgeparavicini, line 21: Not sure if filtering the order is what we mean. We filter the users.

jorgeparavicini (approved):

PR #285, opened 2025-06-20
feat: add login button

yohoji:


LinoGiger (approved):
image
is it possible to move it a bit more to the right? otherwise it looks good

PR #330, opened 2025-07-31
added rapidata config

LinoGiger:


jorgeparavicini (approved):
and how do you configure it?

PR #347, opened 2025-08-14
Feat/rapid 5535 add traces

LinoGiger:


src/rapidata/rapidata_client/order/_rapidata_dataset.py
```diff
@@ -291,17 +305,31 @@ def _process_uploads_in_chunks(
...
+            """Wrapper function that runs _process_single_upload with the provided context."""
+            token = otel_context.attach(context)
+            try:
+                return self._process_single_upload(datapoint, index)
+            finally:
+                otel_context.detach(token)
+
+        # Capture the current OpenTelemetry context before creating threads
+        current_context = otel_context.get_current()
+
```
LucStr, line 320: this looks like the same confusing method from above, still unsure what it does

LinoGiger: same as above

PR #547, opened 2026-04-23
fix(examples): resolve pyright errors outside CI scope

RapidPoseidon:
## Summary

Pyright errors that live in `examples/` and so weren't caught by the `type-check` CI job (which only scopes to `src/rapidata/rapidata_client` per `pyrightconfig.json`). All 4 errors pre-existed the OpenAPI regen — none of them are caused by it.

- `rapi.order.filters.language([...])` → `.Language` in `basic_select_words_order.py`, `select_words_order.py`, `select_faulty_prompt_order.py`. `RapidataFilters` is a class that exposes the filter types as PascalCase class-level aliases (`Country`, `Language`, `Not`, `Or`, `And`).
- Drop unused `from rapidata.rapidata_client.assets._media_asset import MediaAsset` in `conditional_validation_rapid_selection.py` — that module was removed in #316 when assets moved into the `Datapoint` object.

Branched off `openapi-generation-main` (not `main`) so the OpenAPI-regen fix travels with this PR while the upstream branch keeps getting force-pushed.

## Test plan

- [x] `pyright examples` → 0 errors
- [x] `pyright src/rapidata/rapidata_client` (CI scope) → still 0 errors
- [ ] Spot-check that the renamed example calls (`Language(["en"])`) compile against current `RapidataFilters`

🔗 Session: https://session-0cbff43f.poseidon.rapidata.internal/


RapidPoseidon:
Superseded: re-opening against `main` per owner request.

PR #561, opened 2026-04-24
refactor(client): use uuid4 for session id instead of random.Random

RapidPoseidon:
## Summary

\`\`\`python
tracer.set_session_id(
    uuid.UUID(int=random.Random().getrandbits(128), version=4).hex
)
\`\`\`

This reinvents `uuid4` using the Mersenne Twister PRNG. `uuid.uuid4()` is seeded from `os.urandom` on every supported platform and is strictly better here — no behavioral change for telemetry correlation, but removes the unnecessary `random` dependency and the non-CSPRNG path.

Low impact (session IDs are observability metadata, not auth), but zero reason to prefer the hand-rolled version.

## Fix

- `tracer.set_session_id(uuid.uuid4().hex)`
- Drop the `random` import.

## Test plan

- [x] `uv run pyright src/rapidata/rapidata_client` → 0 errors

🔗 Session: <https://session-bc38cc85.poseidon.rapidata.internal/>

### DIFF
--- a/src/rapidata/rapidata_client/rapidata_client.py
+++ b/src/rapidata/rapidata_client/rapidata_client.py
@@ -5,7 +5,6 @@ import requests
 from packaging import version
 from rapidata import __version__
 import uuid
-import random
 from rapidata.service.openapi_service import OpenAPIService
 
 from rapidata.rapidata_client.benchmark.rapidata_benchmark_manager import (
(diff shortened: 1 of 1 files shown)


RapidPoseidon:
Closing on your call — but a quick note for future you in case it's useful:

`uuid.uuid4()` and `random.Random()` (a **fresh** instance — note the constructor, not the module) both seed themselves from `os.urandom` and are unaffected by `random.seed()`. Only the module-level `random.*` functions share the seed the user sets. So functionally `uuid.uuid4().hex` and `uuid.UUID(int=random.Random().getrandbits(128), version=4).hex` behave identically here.

Quick demo:

```
random.seed(42); uuid.uuid4().hex                 → b665547c…
random.seed(42); uuid.uuid4().hex                 → ee11f1a1…   (different ✓)

random.seed(42); random.Random().getrandbits(128) → 5192994…
random.seed(42); random.Random().getrandbits(128) → 3375177…   (different ✓)

random.seed(42); random.getrandbits(128)          → 2523365…
random.seed(42); random.getrandbits(128)          → 2523365…   (same — the bad case)
```

Happy to reopen with just a comment on the existing line if you want the intent (`# independent of user-set random.seed()`) documented.

PR #585, opened 2026-05-08
feat(upload): expose compression config on asset upload

kannwism:
## Summary

Surface the asset service's image-compression knobs through the SDK config. Callers can now opt their client into compression and tune `quality` / `max_dimension` / `library` per process by setting:

```python
from rapidata import rapidata_config, CompressionConfig

rapidata_config.upload.compression = CompressionConfig(enabled=True, quality=70, max_dimension=1024)
```

Any field left as `None` keeps the server-side default, so existing callers see no behaviour change.

## Implementation

- New `CompressionConfig` pydantic model in `config/upload_config.py` with quality/max-dim validators and a `cache_suffix()` helper for cache-key fingerprinting.
- New `compression: CompressionConfig | None` field on `UploadConfig`.
- Public re-export from `rapidata` and `rapidata.rapidata_client` so users can `from rapidata import CompressionConfig`.
- `AssetUploader._upload_file_asset` and `_upload_url_asset` now read `rapidata_config.upload.compression` and pass `compress` / `quality` / `maxdim` / `library` kwargs to `asset_file_post` / `asset_url_post` only when set — call shape is unchanged when compression is not configured (forward-compatible with the current generated client).
- The file and URL cache keys include the compression fingerprint so the same asset uploaded under different settings does not collide on a single cache entry.

## Dependencies

This PR is a wrapper-layer addition; the actual server-side knob is added by [`RapidataAI/rapidata-backend#4164`](https://github.com/RapidataAI/rapidata-backend/pull/4164). The OpenAPI client (`src/rapidata/api_client/`) needs to be regenerated against the deployed asset service for the compression kwargs to actually reach the wire — the existing `update-openapi-client` workflow handles that.

Order of operations:

1. Merge & deploy the backend PR.
2. Run the `update-openapi-client` workflow (or wait for the automated dispatch) to regenerate `asset_api.py` with the new params.
3. Merge this PR.

Until step 2 lands, setting `rapidata_config.upload.compression` produces the configured override on the SDK side but the kwargs are dropped at the OpenAPI client boundary; the request will still succeed and use the server-side defaults.

## Out of scope

- Batched URL uploads via `/asset/batch-upload` — `BatchAssetUploader` does not yet pass `compression` because the input model needs OpenAPI regen first. Follow-up PR after step 2 above. `CompressionConfig`'s docstring calls this out.
- Frontend / dashboard UI for the new knobs.

## Test plan

- [x] `uv run pyright src/rapidata/rapidata_client` — 0 errors, 0 warnings.
- [x] Smoke imports + validators verified locally:
  - default `_compression_kwargs(None)` returns `{}`
  - partial config emits only set fields
  - `quality=0` and `max_dimension=0` raise `ValidationError`
  - assignment to `rapidata_config.upload.compression` round-trips
- [ ] Manual smoke (post-OpenAPI-regen): upload a known-large image with `enabled=True, quality=50` and confirm the returned filename ends in `.webp`.

🔗 Session: https://session-2c7d2568.poseidon.rapidata.internal/

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -49,6 +49,23 @@ logger.info("This will be shown") # (2)!
 | `cacheShards` | `int` | `128` | Number of cache shards for parallel access (immutable) |
 | `batchSize` | `int` | `1000` | Number of URLs per batch (100–5000) |
 | `batchPollInterval` | `float` | `0.5` | Batch polling interval in seconds |
+| `compression` | `CompressionConfig \| None` | `None` | Per-upload image-compression override. `None` defers to the server-side defaults; otherwise sets `enabled` / `quality` / `max_dimension` / `library` for every upload through this client. |
+
+#### Compression override
+
+```python
+from rapidata import rapidata_config, CompressionConfig
+
+# Force the asset service to compress images at quality 70 with a max dimension of 1024px,
+# regardless of the server-side default (which is currently off in production).
+rapidata_config.upload.compression = CompressionConfig(
+    enabled=True,
+    quality=70,
+    max_dimension=1024,
+)
+```
+
+Any field left as `None` falls back to the server-side default. Currently applies to single-asset uploads (`/asset/file` and `/asset/url`); batched URL uploads will pick the override up in a follow-up after the OpenAPI client regenerates.
 
 ## Environment Variables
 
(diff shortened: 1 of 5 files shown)


LinoGiger (approved):
glanced over and seems reasonable. just make sure backend also creates a new asset even if the etag is the same

PR #588, opened 2026-05-11
feat(audience): add filter_global shortcut

RapidPoseidon:
## Summary

Adds \`client.audience.filter_global(filters)\` as a shortcut for the common pattern of "filter the global audience by language / country / demographic". Drops a line and removes the magic-string \`get_audience_by_id("global")\` lookup.

Follow-up to #587. **2 files, 42 lines.**

## Before / after

\`\`\`python
# Before
global_audience = client.audience.get_audience_by_id("global")
arabic_speakers = global_audience.filter([LanguageFilter(["ar"])])

# After
arabic_speakers = client.audience.filter_global([LanguageFilter(["ar"])])
\`\`\`

The shortcut just delegates to \`get_audience_by_id("global").filter(filters)\` — same audience, same payload, two API calls. Kept it as a plain delegation rather than a direct endpoint call so the filter behaviour stays in one place (\`RapidataAudience.filter\`).

## Naming

Went with \`filter_global\` rather than a bare \`.filter(...)\` on the manager — \`client.audience.filter(...)\` reads more like "search my existing audiences" than "create a filtered audience from global", and being explicit about the base avoids that ambiguity.

## Test plan

- [x] \`uv run pyright src/rapidata/rapidata_client\` — 0 errors

🔗 Session: [https://session-0439e1c4.poseidon.rapidata.internal/](https://session-0439e1c4.poseidon.rapidata.internal/)

### DIFF
--- a/docs/audiences.md
+++ b/docs/audiences.md
@@ -233,6 +233,19 @@ combined with `AndFilter` / `OrFilter` / `NotFilter` (or the `&` / `|` /
 `~` operators). Multiple filters passed in the list are combined with
 logical AND.
 
+### Filtering the global audience
+
+The global audience is the most common base — slicing it by language,
+country, or demographic is enough for most jobs. There's a shortcut
+that skips the `get_audience_by_id("global")` step:
+
+```py
+from rapidata import LanguageFilter
+
+arabic_speakers = client.audience.filter_global([LanguageFilter(["ar"])])
+job = arabic_speakers.assign_job(new_job_definition)
+```
+
 ## Next Steps
 
 - Learn about [Classification Jobs](examples/classify_job.md) for categorizing data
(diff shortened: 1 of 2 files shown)


RapidPoseidon:
Usage:

\`\`\`python
from rapidata import RapidataClient, LanguageFilter, CountryFilter, DemographicFilter

client = RapidataClient()

# Single-line filter on the global audience — no get_audience_by_id("global") roundabout.
arabic_speakers = client.audience.filter_global([LanguageFilter(["ar"])])

# Compose filters the same way as before (AND of the list).
arabic_speakers_mena = client.audience.filter_global([
    LanguageFilter(["ar"]),
    CountryFilter(["SA", "AE", "EG"]),
    DemographicFilter("age", ["18-29", "30-39"]),
])

# Returned object is a regular RapidataAudience — works anywhere an audience id does.
job = arabic_speakers.assign_job(job_definition)
benchmark.create_leaderboard(name="…", instruction="…", audience_id=arabic_speakers)
\`\`\`

For non-global bases the existing `.filter(...)` on a `RapidataAudience` instance still applies; this just removes the lookup step for the most common case.

PR #590, opened 2026-05-12
refactor(audience): split filtered audience into its own type

RapidPoseidon:
## Summary

Scopes down what a filtered audience can do, so misuse fails at type-check time instead of at runtime.

- New **`RapidataAudienceBase`** holding the operations valid on any audience: `id`, `name`, `filters`, `assign_job`, `find_jobs`, `delete`.
- **`RapidataAudience`** becomes the dimension-audience subclass. Keeps everything that mutates the qualification pool: `add_classification_example`, `add_compare_example`, `update_filters`, `update_name`, `get_examples`, and the `.filter(...)` method.
- New slim **`RapidataFilteredAudience`** subclass returned by `RapidataAudience.filter(...)`. No extra methods on top of the base — filtered audiences share the base's qualified pool, so adding examples or updating recruitment filters on the slice would be meaningless or actively confusing.
- `RapidataBenchmark.create_leaderboard` widens `audience_id` to `str | RapidataAudienceBase | None` so both audience types pass through.

No behaviour change for existing dimension-audience callers — public methods on `RapidataAudience` keep the same names and signatures.

## Why

Today everything `.filter(...)` returns is a `RapidataAudience`, so this compiles and silently does the wrong thing:

```python
filtered = base.filter([LanguageFilter(["ar"])])
filtered.add_classification_example(...)   # mutates the BASE audience's qualification pool
filtered.filter([CountryFilter(["DE"])])   # nested filtering on a filtered view
filtered.update_filters([...])             # tries to change recruiting filters of a view that doesn't recruit
```

After this PR each of those is a pyright error. Verified locally:

```
$ uv run pyright misuse.py
error: Cannot access attribute "add_classification_example" for class "RapidataFilteredAudience"
error: Cannot access attribute "filter" for class "RapidataFilteredAudience"
```

## Public surface

| Method                       | RapidataAudience | RapidataFilteredAudience |
|------------------------------|:---:|:---:|
| `id`, `name`, `filters`      | ✅ | ✅ |
| `assign_job`                 | ✅ | ✅ |
| `find_jobs`                  | ✅ | ✅ |
| `delete`                     | ✅ | ✅ |
| `filter`                     | ✅ | — |
| `update_filters`             | ✅ | — |
| `update_name`                | ✅ | — |
| `add_classification_example` | ✅ | — |
| `add_compare_example`        | ✅ | — |
| `get_examples`               | ✅ | — |

Both pass `isinstance(x, RapidataAudienceBase)`. Code that previously did `isinstance(x, RapidataAudience)` to mean "dimension audience" continues to work — that's the new intent of the check.

## Not in scope

- `RapidataAudienceManager.get_audience_by_id` still hits `GET /audience/{id}` and returns `RapidataAudience` regardless of id prefix. Resolving an `fau_…` id would need either a backend endpoint change or a `GetAudienceInformation`-style branch on the SDK side — leaving as a separate change.

## Test plan

- [x] `uv run pyright src/rapidata/rapidata_client` — 0 errors, 0 warnings
- [x] Pyright catches `add_classification_example` / `.filter` on a `RapidataFilteredAudience` (verified locally — see above)
- [x] `RapidataFilteredAudience.__mro__` includes `RapidataAudienceBase`
- [x] `RapidataFilteredAudience` public methods = `{name, filters, assign_job, find_jobs, delete}`

🔗 Session: [https://session-0439e1c4.poseidon.rapidata.internal/](https://session-0439e1c4.poseidon.rapidata.internal/)

### DIFF
--- a/src/rapidata/rapidata_client/audience/_audience_base.py
+++ b/src/rapidata/rapidata_client/audience/_audience_base.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from typing import TYPE_CHECKING
-from rapidata.rapidata_client.config import logger, tracer, managed_print
+from rapidata.rapidata_client.config import logger, tracer
 
 if TYPE_CHECKING:
     from rapidata.service.openapi_service import OpenAPIService
@@ -50,16 +50,6 @@ class RapidataAudienceBase:
     def filters(self) -> list[RapidataFilter]:
         """The filters applied to the audience."""
         return self._filters
-
-    def delete(self) -> None:
-        """Deletes the audience."""
-        with tracer.start_as_current_span(f"{type(self).__name__}.delete"):
-            logger.info("Deleting audience '%s'", self)
-            self._openapi_service.audience.audience_api.audience_audience_id_delete(
-                self.id
-            )
-            logger.debug("Audience '%s' has been deleted.", self)
-            managed_print(f"Audience '{self}' has been deleted.")
 
     def assign_job(self, job_definition: RapidataJobDefinition) -> RapidataJob:
         """Assign a job to this audience.
--- a/src/rapidata/rapidata_client/audience/_audience_base.py
+++ b/src/rapidata/rapidata_client/audience/_audience_base.py
@@ -1,0 +1,144 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+from rapidata.rapidata_client.config import logger, tracer, managed_print
+
+if TYPE_CHECKING:
+    from rapidata.service.openapi_service import OpenAPIService
+    from rapidata.rapidata_client.filter import RapidataFilter
+    from rapidata.rapidata_client.job.rapidata_job_definition import (
+        RapidataJobDefinition,
+    )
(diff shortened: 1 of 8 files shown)


src/rapidata/rapidata_client/audience/_audience_base.py
```diff
@@ -0,0 +1,134 @@
...
+                definition_id=job_definition.id,
+                openapi_service=self._openapi_service,
+            )
+            logger.info(f"Assigned job to audience: {self.id}")
+            return job
+
+    def find_jobs(
+        self, name: str = "", amount: int = 10, page: int = 1
+    ) -> list[RapidataJob]:
+        """Find jobs assigned to this audience.
```
LinoGiger, line 94: does this work with filtered audiences?

LinoGiger (approved):

src/rapidata/rapidata_client/audience/_audience_base.py
LucStr: yes it does

PR #598, opened 2026-05-21
feat(signal): add SignalManager to Python SDK

RapidPoseidon:
## Summary

Adds SDK support for the new Signal Service:

- `RapidataSignalManager` exposed as `client.signals` (CRUD entry point: `create`, `get`, `list`, `get_run`).
- `RapidataSignal` with lifecycle methods (`pause`, `resume`, `delete`, `trigger`, `update`), `get_runs` / `get_run`, and a blocking `wait_for_next_run(timeout, poll_interval)` helper that returns the first new terminal run.
- `SignalRun` dataclass with `is_terminal` / `succeeded` convenience properties.

## Implementation

Built on the **autogenerated `SignalApi`** — the signal service is already part of the generated OpenAPI client on `main`, so this PR is just the hand-written manager / entity / run layer plus its wiring. The manager and entity call `openapi_service.signal.signal_api.*` through the same `RapidataApiClient` wrapper as every other manager, so 4xx/5xx are converted to `RapidataError` automatically. (An earlier revision of this branch shipped a raw-HTTP stopgap and regenerated the client itself; both are now obsolete and have been dropped — this branch is rebased onto `main`.)

The public surface follows the real backend contract:

- `create` returns only `{id, nextRunAt}`, so `RapidataSignalManager.create()` fetches the full signal before returning it.
- `trigger` spawns the run asynchronously and returns only the signal id, so `RapidataSignal.trigger()` returns `None`; use `wait_for_next_run()` to observe the resulting run.
- `SignalRun` omits `owner_id` / `owner_mail` / `created_at` — the run payload doesn't carry them.
- `list()` exposes `page` / `page_size` (the endpoint already scopes results to the caller's own + public signals).

## Test plan

- `python -c "from rapidata import RapidataClient; print('OK')"`
- `python -c "from rapidata.rapidata_client.signal import RapidataSignalManager, RapidataSignal, SignalRun; print('OK')"`
- `python -c "from rapidata import RapidataSignal, RapidataSignalManager, SignalRun; print('OK')"`
- `pyright src/rapidata/rapidata_client` -> 0 errors

🔗 Session: https://session-b2a2bd76.poseidon.rapidata.internal/


LinoGiger (changes requested):
this should use the generations and not just send direct api requests

PR #612, opened 2026-06-09
feat(job): expose locate job definition

RapidPoseidon:
## What & why

Exposes the previously-private `_create_locate_job_definition` as a public `client.job.create_locate_job_definition`, so customers can create locate jobs (labelers tap the points in a datapoint that match an instruction) directly from the SDK — and documents it.

This closes the gap raised in two Poseidon sessions: customers were told a locate endpoint exists, but it was private and undocumented in the 3.x docs.

While documenting locate, we also fixed a latent bug across **all** the 3.x example docs: every example created a brand-new empty audience and immediately assigned a job to it. A freshly-created audience has no graduated labelers, so the job never collects responses and `get_results()` never returns — i.e. none of the examples actually ran.

### Locate job (original scope)

- **`rapidata_job_manager.py`** — rename `_create_locate_job_definition` → `create_locate_job_definition` (now public; `mkdocstrings` `filters: ["!^_"]` means only the public name renders in the API reference).
- **`docs/examples/locate_job.md`** — new worked example, wired into `mkdocs.yml` nav.
- **`docs/job_definition_parameters.md`** — locate added to the datapoints table, a dedicated Locate section, and the parameter-availability matrix.
- **`docs/starting_page.md`** — locate added to the "What you can do" landing-page table.
- Docstring opener kept as "With this order…" to stay consistent with the other seven `create_*_job_definition` methods.

### Example docs that actually run (added in this PR)

- **`docs/examples/classify_job.md`** and **`docs/examples/compare_job.md`** — restructured into **Simple / Advanced** content tabs (mirroring the 2.x docs):
  - *Simple* runs immediately against a **curated** audience (Coherence for classify, Alignment for compare) that already has graduated labelers.
  - *Advanced* builds a **custom** audience and trains labelers with qualification examples (`add_classification_example` / `add_compare_example`) before assigning the job.
- **`docs/examples/locate_job.md`** — switched off the dead `create_audience(...)` onto the curated **Global** audience; no Advanced tab (the audience API has no locate qualification examples yet — noted inline).
- **`docs/quickstart.md`** — same fix applied to Step 1 and the Complete Example.
- All curated-audience lookups select the pool by **exact name** — `next(a for a in client.audience.find_audiences("X") if a.name == "X")` — because `find_audiences` does a substring match sorted newest-first, so `[0]` would grab a stale custom audience and re-introduce the empty-audience bug.

## Consistency review (done in-session)

- Naming follows the `create_<type>_job_definition` pattern of the public siblings (classification, compare). ✓
- Shared parameter order matches siblings; correctly omits `data_type` / `confidence_threshold` / `quorum_threshold`, matching the parameter-availability matrix. ✓
- Locate is now public across all three surfaces where it applies (`order`, `validation`, `job`). ✓
- Example/quickstart audience selection is consistent across all four docs pages; verified against the live docs account that Coherence / Alignment / Global resolve and job definitions create. ✓
- `mkdocs build --strict` introduces no new warnings (same 5 pre-existing `mri.md` + `box.py` baseline warnings). ✓

## Note on deployment

`docs.rapidata.ai` is **not** auto-deployed on merge — the `Deploy Documentation` workflow is `workflow_dispatch` only. After merge, someone needs to run it (branch `main`, alias `3.x`, set latest) for these docs + the new API-reference entry to go live.

🔗 Session: https://session-9964dbce.poseidon.rapidata.internal/

### DIFF
--- a/docs/examples/classify_job.md
+++ b/docs/examples/classify_job.md
@@ -2,46 +2,111 @@
 
 To learn about the basics of creating a job, please refer to the [quickstart guide](../quickstart.md).
 
-In this example, we want to rate different images based on a Likert scale to assess how well generated images match their descriptions. The `NoShuffleSetting` setting ensures answer options remain in order since they represent a scale.
+In this example, we rate images on a Likert scale to assess how well generated images match their descriptions. The `NoShuffleSetting` keeps the answer options in order, since they represent a scale.
 
-```python
-from rapidata import RapidataClient, NoShuffleSetting
+=== "Simple"
 
-IMAGE_URLS = [
-    "https://assets.rapidata.ai/tshirt-4o.png",
-    "https://assets.rapidata.ai/tshirt-aurora.jpg",
-    "https://assets.rapidata.ai/teamleader-aurora.jpg",
-]
+    The simple version runs straight away on a **curated** audience — a pre-existing pool of trained labelers — so the job starts collecting responses immediately.
 
-CONTEXTS = ["A t-shirt with the text 'Running on caffeine & dreams'"] * len(IMAGE_URLS)
+    ```python
+    from rapidata import RapidataClient, NoShuffleSetting
 
-client = RapidataClient()
+    IMAGE_URLS = [
+        "https://assets.rapidata.ai/tshirt-4o.png",
+        "https://assets.rapidata.ai/tshirt-aurora.jpg",
+        "https://assets.rapidata.ai/teamleader-aurora.jpg",
+    ]
 
-audience = client.audience.create_audience(name="Likert Scale Audience")
-audience.add_classification_example(
-    instruction="How well does the image match the description?",
-    answer_options=["1: Not at all", "2: A little", "3: Moderately", "4: Very well", "5: Perfectly"],
-    datapoint="https://assets.rapidata.ai/tshirt-4o.png",
-    truth=["5: Perfectly"],
-    context="A t-shirt with the text 'Running on caffeine & dreams'"
-)
+    CONTEXTS = ["A t-shirt with the text 'Running on caffeine & dreams'"] * len(IMAGE_URLS)
 
-job_definition = client.job.create_classification_job_definition(
-    name="Likert Scale Example",
-    instruction="How well does the image match the description?",
-    answer_options=["1: Not at all", "2: A little", "3: Moderately", "4: Very well", "5: Perfectly"],
-    contexts=CONTEXTS,
-    datapoints=IMAGE_URLS,
-    responses_per_datapoint=25,
-    settings=[NoShuffleSetting()] # (1)!
-)
+    client = RapidataClient()
 
-job_definition.preview()
+    audience = next( # (1)!
+        a for a in client.audience.find_audiences("Coherence") if a.name == "Coherence"
+    )
 
-job = audience.assign_job(job_definition)
-job.display_progress_bar()
-results = job.get_results()
-print(results)
-```
+    job_definition = client.job.create_classification_job_definition(
+        name="Likert Scale Example",
+        instruction="How well does the image match the description?",
+        answer_options=["1: Not at all", "2: A little", "3: Moderately", "4: Very well", "5: Perfectly"],
+        contexts=CONTEXTS,
+        datapoints=IMAGE_URLS,
+        responses_per_datapoint=25,
+        settings=[NoShuffleSetting()] # (2)!
+    )
 
-1. Keeps the answer options in their specified order. Without this, options are randomized to reduce bias — but for Likert scales you want them ordered.
+    job_definition.preview()
+
+    job = audience.assign_job(job_definition)
+    job.display_progress_bar()
+    results = job.get_results()
+    print(results)
+    ```
+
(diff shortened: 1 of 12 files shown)


docs/examples/classify_job.md
```diff
@@ -2,46 +2,111 @@
...
+        "https://assets.rapidata.ai/teamleader-aurora.jpg",
+    ]
+
+    CONTEXTS = ["A t-shirt with the text 'Running on caffeine & dreams'"] * len(IMAGE_URLS)
+
+    client = RapidataClient()
+
+    audience = next( # (1)!
+        a for a in client.audience.find_audiences("Coherence") if a.name == "Coherence"
+    )
```
LinoGiger: What is this doing? ^^ why not just either find and take the first or if you wanna be very safe just get by id?

docs/examples/classify_job.md
```diff
@@ -2,46 +2,111 @@
...
+    job.display_progress_bar()
+    results = job.get_results()
+    print(results)
+    ```
+
+    1. Creates a new, empty audience. The `add_classification_example` calls below define who qualifies to join it.
+    2. Qualify labelers on the same UI they'll see in the job. Since the job uses `NoShuffleSetting`, the examples use it too — see [Custom Audiences](../audiences.md#matching-the-job-ui-with-settings).
+
+    !!! warning "Review your qualification examples carefully"
+        Every qualification example and its truth must be reviewed before use. An example with a wrong or ambiguous truth filters out good labelers while letting bad ones through — inverting your quality control. Use only examples with a clear, unambiguous correct answer, and add more than the few shown here for production workloads. See [Custom Audiences](../audiences.md) for the full guide.
```
LinoGiger: this is already clearly stated in the custom audiences - so just a reference is fine i think. but it would be nice to have a warning ABOVE the code to warn that this will take significantly longer as it first has to build an entirely new audience

docs/examples/compare_job.md
```diff
@@ -4,50 +4,127 @@ To learn about the basics of creating a job, please refer to the [quickstart gui
...
+        ["https://assets.rapidata.ai/flux_alien.jpg", "https://assets.rapidata.ai/mj_alien.jpg"],
+        ["https://assets.rapidata.ai/flux_duck.jpg", "https://assets.rapidata.ai/mj_duck.jpg"],
+        ["https://assets.rapidata.ai/flux_book.jpg", "https://assets.rapidata.ai/mj_book.jpg"]
+    ]
+
+    client = RapidataClient()
+
+    audience = next( # (1)!
+        a for a in client.audience.find_audiences("Alignment") if a.name == "Alignment"
+    )
```
LinoGiger: same as above

docs/examples/compare_job.md
```diff
@@ -4,50 +4,127 @@ To learn about the basics of creating a job, please refer to the [quickstart gui
...
+    job = audience.assign_job(job_definition)
+    job.display_progress_bar()
+    results = job.get_results()
+    print(results)
+    ```
+
+    1. Creates a new, empty audience. The `add_compare_example` calls train and filter the labelers who join it.
+
+    !!! warning "Review your qualification examples carefully"
+        Every qualification example and its truth must be reviewed before use. An example with a wrong or ambiguous truth filters out good labelers while letting bad ones through — inverting your quality control. Use only pairs with a clearly better image, and add more than the few shown here for production workloads. See [Custom Audiences](../audiences.md) for the full guide.
```
LinoGiger: same as above

docs/examples/locate_job.md
```diff
@@ -0,0 +1,40 @@
...
+    "https://assets.rapidata.ai/eac11c3e-ad57-402b-90ed-23378d2ff869.jpg",
+    "https://assets.rapidata.ai/04e7e3c6-5554-47ca-bdb2-950e48ac3e6c.jpg",
+    "https://assets.rapidata.ai/91d9913c-b399-47f8-ad19-767798cc951c.jpg",
+]
+
+client = RapidataClient()
+
+audience = next( # (1)!
+    a for a in client.audience.find_audiences("Global Audience") if a.name == "Global Audience"
+)
```
LinoGiger: this is the same syntax nitbpick, but the audience global audience actually just has the id "global" so you can just use that

docs/examples/locate_job.md
```diff
@@ -0,0 +1,40 @@
+# Locate Job Example
+
+To learn about the basics of creating a job, please refer to the [quickstart guide](../quickstart.md).
+
+In a locate job, labelers tap the points in a datapoint that match your instruction. In this example, we ask people to point out visual artifacts in AI-generated images — a common way to find where a generator went wrong.
+
+Locate jobs run on the **global** audience — the broadest pool of labelers, ready to work immediately. Unlike classification and comparison, locate tasks don't support custom qualification examples yet, so there's no custom-audience variant for this job type.
```
LinoGiger: it's not generally a thing that locate has to run on the global audience. you can also create an audience with locate examples

LinoGiger (changes requested):
i think it also makes sense to add the ability to add locate examples to the audience

LinoGiger (changes requested):
i think this is mostly good now, just missing the adding of the locate example. we can maybe also have an advanced example for that?

LinoGiger (approved):

PR #619, opened 2026-06-14
feat(job): expose ranking, free text, select words and draw job definitions

RapidPoseidon:
## Summary

Exposes the four previously-private job-definition methods on `client.job` and adds matching audience qualification-example methods, so the full set of job types is usable from the SDK.

### Code
- **Make four job-definition methods public** in `RapidataJobManager` (drop the `_` prefix): `create_ranking_job_definition`, `create_free_text_job_definition`, `create_select_words_job_definition`, `create_draw_job_definition`. `client.job` now exposes all 7 user-facing definitions (classification, compare, locate, draw, select words, free text, ranking).
- **Add two audience qualification-example methods**: `add_draw_example` and `add_select_words_example` on both `AudienceExampleHandler` and `RapidataAudience`, with input validation, tracing/logging, and `_try_start_recruiting` wrappers matching the existing classify/compare/locate methods.
- Removed a stale `validation_set_id` line from the ranking docstring (the parameter does not exist in the signature).

### Docs
- New example pages: draw, select words, free text, ranking — wired into `mkdocs.yml` nav and `starting_page.md`.
- Extended the parameter-availability matrix in `job_definition_parameters.md` from 3 to 7 job types, adding `sentences`, `comparison_budget_per_ranking`, `responses_per_comparison`, `random_comparisons_ratio`.

## Testing
- All example jobs were run manually by the author and worked.
- Verified: package imports cleanly; all 7 public job-definition methods and both new audience methods are present; the parameter matrix matches the real method signatures cell-for-cell.

🔗 Session: cli-karl-6239291b


LinoGiger (approved):
i assume you have ran and tested all these examples multiple times

PR #622, opened 2026-06-16
feat(datasets): add context-shortening support to the SDK

RapidPoseidon:
## What

Adds **context-shortening** support to the Python SDK. A long datapoint `context` (e.g. a full scene description) is often far more than a single annotator question needs — this lets the SDK tune it down to what's relevant, keeping it within the length the backend accepts.

### 1. `ContextManager` (`client.context`)
- `client.context.shorten_context(context, question) -> str` — single pair.
- `client.context.shorten_contexts([(context, question), ...]) -> list[str]` — batch.

Both call the new batch endpoint `POST /datasets/shorten-context` (request `{ items: [{context, question}, ...] }` → response `{ items: [{shortenedContext}, ...] }`). Results are cached server-side.

### 2. Context-length check in the creation flow
A new `auto_shorten: bool = False` argument is threaded onto the order and job creation calls and enforced centrally in `_create_general_order` / `_create_general_job_definition` (the chokepoints that have both the datapoints **and** the workflow instruction). For each datapoint whose `context` exceeds the limit:
- **default** — logs a clear `logger.warning` that the backend would reject it;
- **`auto_shorten=True`** — shortens the context for the order/job `instruction` (one batched request) and substitutes the result.

The limit (**400 characters**) is not hardcoded as a guess — it mirrors the backend's enforced datapoint context validation in `datasets-service`:
`CreateDatapointCommandValidator` → `RuleFor(x => x.Context).MaximumLength(400)` (the dataset-group context validator enforces the same).

## Backend dependency ⚠️
This depends on the backend contract `POST /datasets/shorten-context` in **rapidata-backend / datasets-service**, which is being added in parallel and is **not deployed yet**. Until it lands, the API layer ships a **hand-written** wrapper (`service/services/context_service.py`) against the agreed contract, with a `TODO` to drop it and regenerate the OpenAPI client once the endpoint is published. The path is kept as a single constant for an easy one-line swap.

**Please confirm the deployed route** before merge: every existing datasets route in the backend is singular (`/dataset/{id}/...`), but the agreed contract here is plural `/datasets/shorten-context`. The constant `_SHORTEN_CONTEXT_PATH` is the single place to adjust if the deployed path differs.

## Scope / known limitation
`auto_shorten` is wired wherever the question is **cleanly reachable at datapoint-build time** — i.e. order and job creation, where the instruction lives on the workflow (`workflow._get_instruction()`).

It is **not** wired into the **flow** path (`RapidataFlow.create_new_flow_batch`): there the context is a single batch-level value and the ranking instruction isn't available at item-creation time, so there's no question to tune against. Rather than invent a workaround, this PR leaves flow untouched and flags it here. If flow contexts should also be shortened, the question source needs to be decided first.

## Validation
- `python -c "from rapidata import RapidataClient, ContextManager"` → OK
- `pyright src/rapidata/rapidata_client` → 0 errors (also checked the changed `service/` files)
- New files are `black`-clean. I did **not** run `black` across the whole package — it wanted to reformat 27 pre-existing files unrelated to this change, so new/changed lines just match the surrounding file style.
- `uv run --group docs mkdocs build` → succeeds; docs updated in `job_definition_parameters.md`.

🔗 Session: node-05f2b913


src/rapidata/rapidata_client/context/_context_length.py
```diff
@@ -0,0 +1,82 @@
...
+        return
+
+    if auto_shorten and not question:
+        # auto_shorten needs the question to tune the context; without it we
+        # can't shorten, so fall back to warning rather than silently proceed.
+        logger.warning(
+            "auto_shorten=True but no question/instruction was available to shorten "
+            "the context against; leaving %d over-long context(s) unchanged.",
+            len(over_limit),
+        )
```
LinoGiger: why is this nullable to begin with?

src/rapidata/rapidata_client/context/context_manager.py
```diff
@@ -0,0 +1,47 @@
...
+
+        Args:
+            context: The (potentially long) context to shorten.
+            question: The question the context will be shown alongside. The
+                context is tuned to what this question needs.
+
+        Returns:
+            The shortened context.
+        """
+        return self.shorten_contexts([(context, question)])[0]
```
LinoGiger, line 41: not used anywhere, pls remove

src/rapidata/rapidata_client/context/_context_length.py
```diff
@@ -0,0 +1,82 @@
...
+# Keep in sync if the backend limit changes.
+MAX_CONTEXT_LENGTH = 400
+
+
+def enforce_context_length(
+    datapoints: list[Datapoint],
+    question: str | None,
+    auto_shorten: bool,
+    context_manager: ContextManager,
+) -> None:
```
LinoGiger: i don't like that this is it's own loose function. since it anyways needs a context manager pass in, just create it as a private method on the context manager, no?

src/rapidata/rapidata_client/context/_context_length.py
```diff
@@ -0,0 +1,82 @@
...
+                "Datapoint %d: shortened context from %d to %d characters.",
+                index,
+                len(datapoint.context),
+                len(new_context),
+            )
+            datapoint.context = new_context
+        return
+
+    for index, datapoint in over_limit:
+        assert datapoint.context is not None
```
LinoGiger: i don't really get it - the context is nullable. at this point it should never be the case anymore to begin with but i don't see why this is here?

LinoGiger (changes requested):
is there a cleaner way to do this without adding an extra parameter? for example with a setting or something?

RapidPoseidon:
Good call — switched the whole thing to a config flag and dropped the `auto_shorten` parameter entirely. It's now `rapidata_config.upload.autoShortenContext` (default `False`), read inside the enforcement code, so no parameter is threaded through any of the 17 order/job creation methods. Net result is ~180 fewer lines. This also matches the existing precedent of `rapidata_config.upload.compression` (a client-side content transform that lives in config rather than per-call args).

Also addressed the inline comments — replies on each thread. Pushed in 801ef9a.

src/rapidata/rapidata_client/context/_context_length.py
RapidPoseidon: Done —  is gone; the logic is now a private method . It reads the auto-shorten flag from config and reuses the manager's own , so no loose function and nothing passed in.

src/rapidata/rapidata_client/context/_context_length.py
RapidPoseidon: The context guard is needed because `Datapoint.context` is genuinely optional (`str | None` — datapoints without any text context are common). The guard is what selects the over-limit ones. I did remove the *redundant* re-checks/asserts further down though: the filter now captures the context value into the tuple, so there are no more `assert context is not None` lines.

src/rapidata/rapidata_client/context/_context_length.py
RapidPoseidon: Removed. That `assert` only existed to re-narrow the type for pyright after the earlier `is not None` filter — you're right it was noise. The over-limit list now carries the (non-null) context value directly, so the assert is gone.

src/rapidata/rapidata_client/context/_context_length.py
RapidPoseidon: Done — `_context_length.py` is gone; the logic is now a private method `ContextManager._enforce_context_length(datapoints, question)`. It reads the auto-shorten flag from config and reuses the manager's own `shorten_contexts`, so there's no loose function and nothing has to be passed in.

src/rapidata/rapidata_client/context/_context_length.py
RapidPoseidon: The guard is needed because `Datapoint.context` is genuinely optional (`str | None` — datapoints with no text context are common), so this is what selects the over-limit ones. I did remove the *redundant* downstream re-checks/asserts: the filter now captures the context value into the tuple, so there are no more `assert context is not None` lines after it.

src/rapidata/rapidata_client/context/_context_length.py
RapidPoseidon: Removed. That `assert` only existed to re-narrow the type for pyright after the earlier `is not None` filter — you're right it was noise. The over-limit list now carries the (non-null) context value directly, so the assert is gone.

src/rapidata/rapidata_client/context/context_manager.py
RapidPoseidon: This is the single-pair public entry point (`shorten_context(context, question)`) — the feature spec asked for both a single and a batch form of the public API. It's a one-line delegate to `shorten_contexts` so a caller shortening one context doesn't have to wrap it in a list. Happy to drop it if you'd rather keep the surface to the batch method only — let me know and I'll remove it.

src/rapidata/service/services/context_service.py
```diff
@@ -0,0 +1,80 @@
...
+    from rapidata.rapidata_client.api.rapidata_api_client import RapidataApiClient
+
+
+# TODO: Replace this hand-written wrapper with the generated OpenAPI client once
+# `POST /datasets/shorten-context` ships in datasets-service and the contract is
+# published. At that point this service should call the generated
+# ``DatasetsApi`` method (and the request/response shapes below become typed
+# models) instead of crafting the request by hand. The path is kept as a single
+# constant so the regeneration is a one-line removal.
+_SHORTEN_CONTEXT_PATH = "/datasets/shorten-context"
```
LinoGiger: there is the "ContextShorteningApi" that you put under the "dataset service in the api

src/rapidata/service/openapi_service.py
```diff
@@ -77,6 +78,7 @@ def __init__(
         self._leaderboard: LeaderboardService | None = None
         self._rapid: RapidService | None = None
         self._translation: TranslationService | None = None
+        self._context: ContextService | None = None
```
LinoGiger: no it's under the datasetService

LinoGiger (changes requested):

src/rapidata/service/services/context_service.py
RapidPoseidon: Done — switched to the generated `ContextShorteningApi`. Merged main in (the endpoint shipped and the client was regenerated) and deleted the hand-written `ContextService` wrapper entirely. `ContextManager` now calls `dataset.context_shortening_api.datasets_shorten_context_post(...)` with the generated `ShortenContextEndpointInput` / `ShortenContextEndpointInputItem` models. The earlier route concern is moot too — the generated client confirms `/datasets/shorten-context`.

src/rapidata/service/openapi_service.py
RapidPoseidon: Fixed — removed the `OpenAPIService.context` property. The generated `ContextShorteningApi` is now exposed under `DatasetService` as `dataset.context_shortening_api`, alongside `dataset_api` / `datapoints_api`.

PR #625, opened 2026-06-18
fix: resolve benchmark prompt assets without asserting concrete asset types

RapidPoseidon:
## Problem

`RapidataBenchmark.identifiers` (and `prompts` / `prompt_assets` / `tags`) crashes on **any benchmark whose prompts have file assets** — which blocks `add_model` and `add_prompts`, since both read `self.identifiers` internally.

`__instantiate_prompts` did:

```python
assert isinstance(prompt.prompt_asset.actual_instance, FileAssetModel)
source_url = prompt.prompt_asset.actual_instance.metadata["sourceUrl"].actual_instance
assert isinstance(source_url, SourceUrlMetadataModel)
```

Neither assumption holds against the current API response:

1. `prompt.prompt_asset` is an **`IAssetModel` oneOf wrapper**; its `.actual_instance` is an `IAssetModel*` variant (`IAssetModelFileAssetModel` / `…MultiAssetModel` / `…TextAssetModel` / `…NullAssetModel`) — never the bare `FileAssetModel`. → `AssertionError`.
2. `metadata` is `Dict[str, IMetadataModel]` and only carries a `"sourceUrl"` entry for **URL-registered** assets. File-uploaded prompts only have `originalFilename` / `imageDimension`, so `metadata["sourceUrl"]` would `KeyError`.
3. The metadata variant is `IMetadataModelSourceUrlMetadataModel`, not the bare `SourceUrlMetadataModel`, so the second assert is wrong too.

Reproduced live against a real file-asset benchmark (499 prompts): `benchmark.identifiers` raised `AssertionError` before this change.

## Fix

Resolve the source URL defensively instead of asserting concrete types, mirroring the existing pattern in `RapidataFlowItem._extract_asset_key()`:

```python
@staticmethod
def __source_url(asset):
    instance = getattr(asset, "actual_instance", None)
    metadata = getattr(instance, "metadata", None)
    if not metadata:
        return None
    source_url = metadata.get("sourceUrl")
    actual = getattr(source_url, "actual_instance", None)
    return getattr(actual, "url", None)
```

`prompt_assets` now yields the source URL when present, else `None` for file-uploaded assets that have no public URL in the response — the old code could never have returned a URL for those anyway; it crashed.

## Verification

- Live against the real benchmark (read-only, creates nothing): `identifiers`, `prompts`, `prompt_assets`, `tags` all populate (499 each) with no exception; `add_model`'s identifier-membership check passes.
- `uv run black src/rapidata/rapidata_client` → unchanged.
- `uv run pyright src/rapidata/rapidata_client/benchmark/rapidata_benchmark.py` → 0 errors, 0 warnings.

## Notes

- **No version bump** — versioning/publishing is CI-automated (`release_and_publish.yml`).
- **No test added** — the repo has no test framework yet; a regression test for this (`IAssetModel*` wrapper + missing `sourceUrl`) would be worth adding once test infra exists.
- Discovered while adding participants to a mock benchmark in `Image-Editing-Benchmarks` (RapidataAI/Image-Editing-Benchmarks#6), which currently ships a consumer-side cache-injection workaround that can be removed once this releases.

🔗 Session: ${POSEIDON_SESSION_URL:-$HOSTNAME}


LinoGiger:
had to fix it in a different way: https://github.com/RapidataAI/rapidata-python-sdk/pull/634

RapidPoseidon:
Thanks @LinoGiger — closing this in favor of #634, which is merged and is the better fix: falling back to `originalFilename` (so `prompt_assets` returns a value for file-uploaded assets instead of my `None`) and normalizing the cached value so it stays consistent across a re-fetch are both improvements over what I had here. Same root cause (the `IAssetModel`/`IMetadataModel` oneOf wrappers vs. the bare `FileAssetModel`/`SourceUrlMetadataModel`), better handling.

For traceability: this surfaced while adding participants to a benchmark with image prompts (`add_model` → `identifiers` crashed). The downstream repo (RapidataAI/Image-Editing-Benchmarks#6, merged) carries a temporary cache-injection workaround; I'll drop it once a release containing #634 is published (not tagged yet as of this writing).

PR #626, opened 2026-06-18
feat(benchmark): add participant deletion and elo lookup

RapidPoseidon:
## What

Adds two capabilities to `BenchmarkParticipant` (the participant/model of a benchmark) in the Python SDK:

- **`participant.delete()`** — removes the participant and its uploaded media from the benchmark (calls `participant/{id}` DELETE).
- **`participant.get_elo()`** — returns the participant's current Elo score, aggregated across all of the benchmark's leaderboards, or `None` when it hasn't been computed yet.

## How

`get_elo()` queries the benchmark standings filtered to this participant (`benchmark/{id}/standings?participantIds=...`) and returns the `score`. To do this without an extra round-trip, the participant now carries its `benchmark_id`, threaded in from the two construction sites in `RapidataBenchmark` (`participants` property and `add_model`).

Docs updated under `docs/mri_advanced.md` (new "Inspecting a Participant's Elo" and "Deleting Participants" sections).

## Usage

```python
participant = benchmark.participants[0]
print(participant.get_elo())  # e.g. 1043.2, or None
participant.delete()
```

## Checks

- `pyright src/rapidata/rapidata_client` → 0 errors
- New code is black-formatted

🔗 Session: ${POSEIDON_SESSION_URL:-$HOSTNAME}

### DIFF
--- a/src/rapidata/rapidata_client/benchmark/participant/participant.py
+++ b/src/rapidata/rapidata_client/benchmark/participant/participant.py
@@ -63,10 +63,12 @@ class BenchmarkParticipant:
             The Elo score, or ``None`` if it has not been computed yet (for
             example when the participant has not been evaluated).
         """
+        # The full standings must be requested: scores are recomputed relative
+        # to the whole field on every call, so filtering to a single participant
+        # would yield a meaningless score.
         with tracer.start_as_current_span("BenchmarkParticipant.get_elo"):
             result = self._openapi_service.leaderboard.benchmark_api.benchmark_benchmark_id_standings_get(
                 benchmark_id=self._benchmark_id,
-                participant_ids=[self.id],
             )
 
             for standing in result.items:
--- a/src/rapidata/rapidata_client/benchmark/participant/participant.py
+++ b/src/rapidata/rapidata_client/benchmark/participant/participant.py
@@ -5,7 +5,7 @@ import time
 from typing import Literal
 from tqdm.auto import tqdm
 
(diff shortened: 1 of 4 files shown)


src/rapidata/rapidata_client/benchmark/participant/participant.py
```diff
@@ -50,6 +53,39 @@ def status(self) -> ParticipantStatus:
...
+
+        Returns:
+            The Elo score, or ``None`` if it has not been computed yet (for
+            example when the participant has not been evaluated).
+        """
+        with tracer.start_as_current_span("BenchmarkParticipant.get_elo"):
+            result = self._openapi_service.leaderboard.benchmark_api.benchmark_benchmark_id_standings_get(
+                benchmark_id=self._benchmark_id,
+                participant_ids=[self.id],
+            )
```
LinoGiger, line 72: this doesn't work. it will request the standings for only 1 participant which doesn't work. because the scores get recalculated every time. so you have to request the full results and then pick the elo of the current one.

LinoGiger (changes requested):

src/rapidata/rapidata_client/benchmark/participant/participant.py
RapidPoseidon: Good catch — fixed in a5b1119. Removed the `participant_ids` filter so `get_elo()` now fetches the full benchmark standings and picks out this participant's score, instead of asking the server to recompute Elo on a single-participant field.

PR #655, opened 2026-07-08
feat(auth): support shared token file for distributed workers

RapidPoseidon:
## Why

A customer running the SDK (ranking flows) across thousands of GPU workers hit auth-scale problems: every worker's `RapidataClient()` exchanged the client credentials for its own token at startup, and since all tokens expire together (~1 h), all workers re-hit the identity service in the same instant. They worked around it with one coordinator process that refreshes the token and writes it to a shared file, with workers reusing it.

The SDK already half-supports this via the `token=` dict param — but a pre-supplied client-credentials token has no refresh path in authlib (no `refresh_token`, no `grant_type` metadata), so once it expires every request raises `InvalidTokenError` and the client must be rebuilt.

## What

- **`RapidataClient(token_file=...)`** (or `RAPIDATA_TOKEN_FILE` env var): the client loads the token JSON from the file at startup and **re-reads the file whenever the in-memory token is within `leeway` (60 s) of expiry**. Workers never hold the client secret and never need to be reconstructed — the coordinator just keeps the file fresh.
- Plumbed through `OpenAPIService` → `RESTClientObject.setup_oauth_with_token` (both `rest.py` and its `rest.mustache` template, per repo convention).
- New docs page **Distributed Training** under the Ranking Flows section documenting the full pattern: coordinator loop (absolute `expires_at`, atomic temp-file + `os.replace` writes), worker usage, startup-jitter tip, and the recreate-the-client fallback for older SDK versions. Cross-linked from the Authentication guide.

## Verification

- `pyright src/rapidata/rapidata_client src/rapidata/service`: 0 errors
- `black --check` clean on touched files under `src/rapidata/rapidata_client`
- Functional smoke test: init from an expired token file → first request path reloads the fresh token from the file; valid in-memory token → no reload; `RapidataClient()` construction via `RAPIDATA_TOKEN_FILE` alone (no client credentials) works
- `uv run --group docs mkdocs build` succeeds; new page renders in nav

🔗 Session: https://session-43887903.poseidon.rapidata.internal/

### DIFF
--- a/docs/distributed_training.md
+++ b/docs/distributed_training.md
@@ -36,37 +36,20 @@ restarted or re-authenticated.
 
 ## Coordinator
 
-The coordinator exchanges the client credentials for a token (see
-[Authentication](authentication.md)) and rewrites the shared file ahead of
-expiry:
+The coordinator authenticates with the client credentials as usual (see
+[Authentication](authentication.md)), exports its token with `get_token()`,
+and rewrites the shared file ahead of expiry:
 
 ```python
 import json
 import os
 import time
 
-import requests
+from rapidata import RapidataClient
 
-TOKEN_URL = "https://auth.rapidata.ai/connect/token"
 TOKEN_FILE = "/shared/rapidata_token.json"
 
-
-def fetch_token() -> dict:
-    response = requests.post(
-        TOKEN_URL,
-        data={
-            "grant_type": "client_credentials",
-            "client_id": os.environ["RAPIDATA_CLIENT_ID"],
-            "client_secret": os.environ["RAPIDATA_CLIENT_SECRET"],
-            "scope": "openid roles email api",
-        },
-    )
-    response.raise_for_status()
-    token = response.json()
-    # Workers decide when to re-read the file based on this absolute
-    # timestamp — the relative expires_in is meaningless to a later reader.
-    token["expires_at"] = int(time.time()) + token["expires_in"]
-    return token
+coordinator = RapidataClient()  # (1)!
 
 
 def write_token(token: dict) -> None:
@@ -79,18 +62,24 @@ def write_token(token: dict) -> None:
 
 
 while True:
-    token = fetch_token()
+    token = coordinator.get_token()  # (2)!
     write_token(token)
-    # Refresh 5 minutes before expiry so workers never read a stale token.
-    time.sleep(max(token["expires_in"] - 300, 60))
+    # Rewrite 5 minutes before expiry so workers never read a stale token.
+    time.sleep(max(token["expires_at"] - time.time() - 300, 60))
 ```
 
-Two details matter:
+1. The coordinator is the only process that holds the client credentials —
+   via `RAPIDATA_CLIENT_ID` / `RAPIDATA_CLIENT_SECRET` or the cached
+   credentials file.
+2. Returns the complete token object, refreshing it first if it has
+   expired. It includes the absolute `expires_at` timestamp workers use to
+   decide when to re-read the file.
 
-1. **Write an absolute `expires_at`** (Unix timestamp, seconds). Without it
-   the SDK cannot tell that a token read from the file is about to expire.
-2. **Write atomically** (temp file + `os.replace`), so workers never parse a
-   half-written file.
+**Write the file atomically** (temp file + `os.replace`, as above) so
+workers never parse a half-written file. If you produce the token without
+the SDK (e.g. a `curl` call to the token endpoint), you must also add the
+absolute `expires_at` yourself — the raw response only carries the relative
+`expires_in`, which is meaningless to a later reader.
 
 ## Practical tips
 
--- a/docs/distributed_training.md
+++ b/docs/distributed_training.md
@@ -1,0 +1,120 @@
+# Distributed Training
+
+When you use Rapidata from a large distributed job — for example a ranking
+flow queried from hundreds or thousands of GPU workers — don't let every
+worker authenticate on its own. Each `RapidataClient()` exchanges the client
+credentials for an access token at startup, and because all those tokens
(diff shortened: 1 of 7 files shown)


docs/distributed_training.md
```diff
@@ -0,0 +1,120 @@
...
+            "client_secret": os.environ["RAPIDATA_CLIENT_SECRET"],
+            "scope": "openid roles email api",
+        },
+    )
+    response.raise_for_status()
+    token = response.json()
+    # Workers decide when to re-read the file based on this absolute
+    # timestamp — the relative expires_in is meaningless to a later reader.
+    token["expires_at"] = int(time.time()) + token["expires_in"]
+    return token
```
LinoGiger: should this not use the sdk client methods?

RapidPoseidon: Good call — the SDK had no public way to get the token out, which is why the example went to the raw endpoint. Added `RapidataClient.get_token()` in 2efaa77: it returns the complete token object (with absolute `expires_at`), refreshing it first if expired, so the coordinator example is now fully SDK-native. Verified the refresh path: authlib stores `grant_type=client_credentials` in the session metadata on the initial fetch, so `ensure_active_token` re-fetches on demand.

docs/distributed_training.md
```diff
@@ -0,0 +1,109 @@
...
+    # token or the new one, never a partially written file.
+    os.replace(tmp_path, TOKEN_FILE)
+
+
+while True:
+    token = coordinator.get_token()  # (2)!
+    write_token(token)
+    # Rewrite 5 minutes before expiry so workers never read a stale token.
+    time.sleep(max(token["expires_at"] - time.time() - 300, 60))
+```
```
LinoGiger: can you swap the coordinator and worker around, since the coordinator has to be started first, i think it makes more sense? 

additionally add a os.makedir for the /shared and add a note that in this example it's gonna be added as a root file

RapidPoseidon: Done in 6a637e4 — coordinator section now comes first, the example creates the directory with `os.makedirs(..., exist_ok=True)`, and an annotation notes that the root-level `/shared` is just for the example (point it at whatever storage the workers actually mount).

Your sleep(1) test also surfaced something worth fixing: `get_token()` intentionally returns the cached token while it's valid and only refreshes within `leeway` of expiry (refreshing every call would be the stampede all over again) — but `leeway` was silently ignored on the client-credentials path, so the example's '5 minutes early' promise couldn't actually hold. Now plumbed through: the coordinator example uses `RapidataClient(leeway=300)` + a simple `time.sleep(60)` loop, which renews the file 5 minutes before expiry.

docs/distributed_training.md
```diff
@@ -0,0 +1,116 @@
...
+    with open(tmp_path, "w") as f:
+        json.dump(token, f)
+    # Atomic replace: a worker reading concurrently sees either the old
+    # token or the new one, never a partially written file.
+    os.replace(tmp_path, TOKEN_FILE)
+
+
+while True:
+    write_token(coordinator.get_token())  # (3)!
+    time.sleep(60)
```
jorgeparavicini, line 85: I'm wondering whether we want to implement this as well. I mean it's quite boilerplatey. We could have a method that can take a path and a duration and we would continously write to it. If the customer would want to write the token differently they could still write this part themselves. But i think having a default implementation would make adoption of this pattern easier

docs/distributed_training.md
```diff
@@ -0,0 +1,116 @@
...
+   path, a shared volume, …).
+2. The coordinator is the only process that holds the client credentials —
+   via `RAPIDATA_CLIENT_ID` / `RAPIDATA_CLIENT_SECRET` or the cached
+   credentials file. `leeway=300` treats the token as expired once it is
+   within 5 minutes of its actual expiry, so the file is renewed well before
+   workers need it.
+3. `get_token()` returns the current token — including the absolute
+   `expires_at` timestamp workers rely on — and only fetches a new one when
+   it is within `leeway` of expiry. Polling it every minute therefore does
+   **not** spam the auth server; most iterations just rewrite the same token.
```
jorgeparavicini: It's also important to mention that doing this exposes the token and the path where the token is written to should be kept secure from unauthorized access. Yes it's obvious but its important to state the obviousness in these sitatuiosn

jorgeparavicini: This is an implementation detail that only concerns us, not our customers. This documentation should be written with details useful for the customer. The point it makes is valid but it can use some rewording

docs/distributed_training.md
RapidPoseidon: Agreed — added `RapidataClient.maintain_token_file(path, interval=60)` in 36c887e. It writes the file once synchronously (so it's guaranteed to exist — or fail loudly — before workers start), creates the directory, then keeps rewriting it atomically from a daemon thread; transient failures inside the loop are logged and retried on the next interval instead of killing the thread. It returns the thread so a dedicated coordinator can just `.join()` it, while a rank-0-that-also-trains drops the join. The manual loop moved to a 'Writing the file yourself' subsection for custom setups.

docs/distributed_training.md
RapidPoseidon: Added a dedicated warning admonition in 36c887e: the file contains a bearer token, anyone who can read it can act on the account until expiry, so write it only to storage the training job alone can access and restrict permissions. Also noted in the `maintain_token_file` docstring.

RapidPoseidon: Reworded in 36c887e — the annotation now frames it from the customer's side: `get_token()` is cheap to call at any frequency; it returns the current token and only contacts the auth server once the token is within `leeway` of expiry. (With `maintain_token_file` as the headline API, the manual loop and this note now live in the 'Writing the file yourself' subsection.)

jorgeparavicini (approved):

PR #671, opened 2026-07-12
fix(job): surface review/spend-limited states and cost warning

RapidPoseidon:
## What & why

Workstream E (Wave 2) of the RFC *Job submission: cost warning + transparent review outcomes*. Two customer-facing SDK gaps:

1. **Silent hang.** `RapidataJob._wait_for_status` waited only for `Completed`/`Failed`, so a job that entered `ManualApproval` or `SpendLimited` blocked the caller **forever** (most visibly via `get_results()`). It now raises an informative error instead:
   - **ManualApproval** → names the review reason when the API returns one (`reviewReason`), and falls back to a generic *"is being reviewed"* message when it's null (a job can legitimately be under review with no reason yet).
   - **SpendLimited** → says partial results remain available and the account needs a top-up to resume.
   - `display_progress_bar` had the identical loop and gets the same guard, so the two entry points behave consistently.

2. **Cost warning.** When job creation returns the new `costWarning { estimatedCost, availableBalance, shortfall }`, `assign_job` now emits a Python `warnings.warn` (estimate, balance, shortfall), phrased as an estimate. The job is still created and behavior is otherwise unchanged — absent `costWarning` → no warning.

Follow-up commit folds in code-review findings: the stale "blocks until complete" promise in `quickstart.md` now points at the new review/cost-warning docs; the polling loops fetch the job GET output once per iteration instead of re-fetching it for the review reason / failure message when raising; and the blocking-state comparisons use the generated `AudienceJobState` enum instead of raw strings (matching how `RapidataOrder` uses `OrderState`).

## Tests

Introduces `pytest` (dev dep) and `tests/test_job_review_and_cost_warning.py`:
- hang fix on `ManualApproval` **with** and **without** a `reviewReason`
- hang fix on `SpendLimited` (+ `display_progress_bar`)
- `costWarning` emission and the no-warning path
- unchanged happy path (`Completed` → results returned)

`pyright`, `black`, `pytest` (7/7), and `mkdocs build` all clean; docs notes in `audiences.md` and `quickstart.md`.

Depends on the generated models from **Update OpenAPI Schemas** #670 — merged, and this branch is now rebased on top of it, so the diff is only the hand-written changes.

Note: no CI job runs pytest today (CI is pyright-only); wiring the new suite into CI is a sensible follow-up.

🔗 Session: https://node-07688719.poseidon.rapidata.internal/
🔗 Review session: https://cli-karl-57b3b766.poseidon.rapidata.internal/


src/rapidata/rapidata_client/audience/_audience_base.py
```diff
@@ -89,8 +93,29 @@ def assign_job(self, job_definition: RapidataJobDefinition) -> RapidataJob:
...
+        if cost_warning is None:
+            return
+        warnings.warn(
+            f"Job '{job.name}' has an estimated cost of {cost_warning.estimated_cost:.2f}, "
+            f"but the account balance is {cost_warning.available_balance:.2f} — it will "
+            f"likely pause about {cost_warning.shortfall:.2f} short of finishing. The job "
+            f"was created and runs as far as the balance allows; top up the account to let "
+            f"it complete. This is an estimate.",
+            stacklevel=3,
+        )
```
LinoGiger: why does this not just use a logger?

tests/test_job_review_and_cost_warning.py
```diff
@@ -0,0 +1,150 @@
+"""Tests for surfacing job review/spend-limited states and the create cost warning.
```
LinoGiger: since when do we have test files on the sdk? like i don't mind it in general. but please create a better folder structure than this ^^

LinoGiger (changes requested):

RapidPoseidon:
Thanks @LinoGiger — both addressed in `7ffa270` (on top of your `535d8d6` polling refactor, which I kept):

1. **Logger, not `warnings.warn`.** `assign_job` now surfaces the cost warning via `logger.warning(...)` (lazy `%`-args), matching the SDK's existing user-facing warning convention (e.g. `rapidata_client.py`). Docs wording updated accordingly.
2. **Test folder structure.** Split the flat file into a package layout mirroring `src/`:
   - `tests/rapidata_client/job/test_rapidata_job.py` — hang fixes + `display_progress_bar` + happy path
   - `tests/rapidata_client/audience/test_audience_base.py` — cost-warning emit / no-emit

   Kept your `AudienceJobState(state)` test-helper change so the enum-based blocking checks are exercised correctly.

`pyright`, `black`, `pytest` (7 passing) and `mkdocs build` all green locally.

PR #674, opened 2026-07-13
feat(audience): expose device filter for filtered audiences

RapidPoseidon:
## What

Enables `DeviceFilter` when deriving a **filtered audience** — the SDK follow-up to backend [rapidata-backend#4697](https://github.com/RapidataAI/rapidata-backend/pull/4697), now that the audience device-filter model is in the generated client.

## Why

`DeviceFilter` worked for orders/jobs but `._to_audience_model()` raised `NotImplementedError`, so `RapidataAudience.filter([DeviceFilter(...)])` failed client-side. The backend now accepts a device audience filter, so the SDK can emit it.

## How

- `DeviceFilter._to_audience_model()` now returns `IAudienceFilter(IAudienceFilterDeviceAudienceFilter(_t="DeviceFilter", deviceTypes=...))` (mirrors the existing `_to_model()` order path and the other `_to_audience_model()` filters).
- `BackendFilterMapper` decodes `IAudienceFilterDeviceAudienceFilter` back into a `DeviceFilter`, so reading an audience's filters round-trips.
- Docs (`audiences.md`) and the `filter()` docstring list `DeviceFilter` as supported — and I added the previously-missing `DemographicFilter` row while there.

## Verification

- `pyright src/rapidata/rapidata_client` — 0 errors.
- `black` — clean.
- `mkdocs build` — succeeds.
- Manual encode→decode round-trip: `DeviceFilter([PHONE, TABLET])` → audience model → back to an equal `DeviceFilter`.

Usage:
```python
from rapidata import DeviceFilter, DeviceType
slice = base.filter([DeviceFilter([DeviceType.PHONE])])
```

🔗 Session: https://session-35b46b6a.poseidon.rapidata.internal/


docs/audiences.md
```diff
@@ -244,6 +244,8 @@ a way that's better expressed as a single combined filter on the base.
 |---|---|
 | `CountryFilter` | ISO-3166 country code (e.g. `["US", "CA"]`) |
 | `LanguageFilter` | Spoken / device language (e.g. `["en", "de"]`) |
+| `DemographicFilter` | Demographic attribute (e.g. `DemographicFilter("age", ["18-29"])`) |
```
LinoGiger: why did we expose this as well?

LinoGiger: if demographic filters are indeed already supported. then please at least put them in an enum

LinoGiger (approved):

LinoGiger (changes requested):
didn't mean to approve

RapidPoseidon:
Good call — the `DemographicFilter` row was an unrelated drive-by addition (the table was missing it, but the docstring already listed it). I've dropped it so this PR stays scoped to the device filter. `DeviceFilter` itself uses the `DeviceType` enum already.

On the underlying point: demographic identifiers/values are currently free strings (`DemographicFilter("age", ["18-29"])`), and turning those into an enum is a worthwhile but separate change with its own design (which identifiers/buckets are valid). Happy to open a follow-up for that if you'd like.

RapidPoseidon:
Done — addressed both comments:

1. **Demographic identifiers are now an enum.** Added `DemographicIdentifier` (`AGE` / `GENDER` / `OCCUPATION`); `DemographicFilter` now takes it instead of a raw string. This was well worth it — the occupation identifier is actually `"occupation-2026-07-01"` (versioned), which no one would type correctly by hand.
2. Re-added `DemographicFilter` to the docs table, now that it's type-safe, with the enum example.

`DeviceFilter` already used the `DeviceType` enum. pyright clean, encode/decode round-trips verified for both filters.

Note this is a **breaking change** for any caller currently passing a string (`DemographicFilter("age", ...)` → `DemographicFilter(DemographicIdentifier.AGE, ...)`).

docs/audiences.md
```diff
@@ -244,6 +244,8 @@ a way that's better expressed as a single combined filter on the base.
 |---|---|
 | `CountryFilter` | ISO-3166 country code (e.g. `["US", "CA"]`) |
 | `LanguageFilter` | Spoken / device language (e.g. `["en", "de"]`) |
+| `DemographicFilter` | Demographic attribute (e.g. `DemographicFilter(DemographicIdentifier.AGE, ["18-29"])`) |
```
LinoGiger: no but i want the whole thing to be enum - so also the 2nd values since the ranges are predefined in the backend

LinoGiger (changes requested):

RapidPoseidon:
Reworked per your call — age and gender only, fully enum:

- `AgeFilter` and `GenderFilter` now work for filtered audiences (they already take the `AgeGroup` / `Gender` enums, whose values match the backend demographic strings exactly). E.g. `base.filter([AgeFilter([AgeGroup.BETWEEN_18_29]), GenderFilter([Gender.FEMALE])])`.
- Removed the free-string `DemographicFilter` and the `DemographicIdentifier` enum entirely. `RapidataFilters` now exposes `Age` / `Gender` instead of `Demographic`.
- **Occupation is deferred** — its value set isn't defined in `rapidata-backend` (only the versioned `occupation-2026-07-01` identifier), so it needs its own enum + source later.

pyright clean, encode/decode round-trips verified for age, gender, and device. Device filter (this PR's original purpose) is unchanged.

PR #690, opened 2026-07-20
feat(audience): explicit recruiting lifecycle + local instruction length validation

RapidPoseidon:
## What & why

Two changes from SDK integration feedback (SDK 3.16.5, Compare task).

### 1. Explicit audience recruiting lifecycle ⚠️ BREAKING

Previously, adding the **first** qualification example to a custom audience implicitly started recruiting: every `RapidataAudience.add_*_example(...)` called the private `_try_start_recruiting()`, which POSTed to the recruit endpoint on the first example. This meant recruiting could begin before all examples were uploaded and reviewed, and a mid-upload failure could leave a partially-configured **live** audience.

Recruiting is now **explicit**:

- Removed the implicit `_try_start_recruiting()` call from every `add_*_example` method and from `_add_rapid_example`.
- Replaced the private helper with a public, chainable **`start_recruiting() -> RapidataAudience`**. It is idempotent per instance (a second call is a no-op via the existing `_recruiting_started` guard), and — unlike the old private version — it **no longer swallows `RapidataError`**, so a genuine failure propagates and the caller knows recruiting did not start.

New lifecycle: `create_audience(...)` → `add_*_example(...)` (no recruiting) → `start_recruiting()`. The backend audience stays in its existing `Created` status until `start_recruiting()` is called; no backend change is involved for this part.

**🚨 Breaking change:** code relying on auto-recruit-on-first-example must now call `audience.start_recruiting()` after all qualification examples are added and reviewed. Without it, the audience never recruits.

### 2. Local instruction length validation (uniform 250)

Instructions over the limit were only rejected by the backend mid-mutation (after assets/datasets were already uploaded), surfacing as a generic `RapidataError`. This validates locally at object construction and raises immediately with a clear message:

```
instruction is 256 characters; maximum is 250
```

- Added `Workflow.MAX_INSTRUCTION_LENGTH = 250` and a shared `validate_instruction_length(...)` helper (single source of truth) on the workflow base class.
- Called it in `__init__` of every instruction-bearing workflow: compare, classify, free text, locate (target), draw (target), select words, multi-ranking, timestamp. (Evaluation workflow has no user-facing instruction field, so nothing to validate.)
- Applied the **same** 250 check to audience qualification examples (which write criteria/title/target directly without building a Workflow) in `audience_example_handler.py` — classification (title), compare (criteria), locate (target), draw (target), select words (title) — **before any asset upload**, reusing the same shared validator so the number isn't duplicated.

The backend limit has been unified to 250 across all SDK-reachable task types; the sibling backend PR lowers Compare/FreeText/Locate/Line from 255 to 250 (Classify/Transcription were already 250).

**Sibling PR:** RapidataAI/rapidata-backend#4798 — `fix(rapid): unify instruction length limit 250`

### 3. Docs & examples

Because recruiting is no longer automatic, every doc that creates a custom audience, adds examples, then runs a job now calls `start_recruiting()` before assigning the job: `docs/audiences.md` (new explicit "Start Recruiting" step + prose note + complete example), `docs/human_prompting.md`, `docs/confidence_stopping.md` (both snippets), and the custom-audience sections of the compare/classify/locate/draw/select-words example docs.

## Verification

- `pyright src/rapidata/rapidata_client` → 0 errors
- `black --check` on touched files → clean
- `from rapidata import RapidataClient` → OK
- `mkdocs build` (docs group) → builds (only pre-existing warnings)
- Boundary check: 250 chars accepted, 256 raises `instruction is 256 characters; maximum is 250`

🔗 Session: https://node-6fff3183.poseidon.rapidata.internal/

### DIFF
--- a/docs/audiences.md
+++ b/docs/audiences.md
@@ -79,22 +79,20 @@ for prompt, datapoint in zip(PROMPTS, DATAPOINTS):
 
 ### Step 3: Start Recruiting
 
-Adding examples no longer starts recruiting on its own. Once all your
-qualification examples are added **and reviewed**, call `start_recruiting()`
-to begin the distilling/onboarding campaign that qualifies labelers against
-them. Until you call it, the audience stays in its `Created` state and
-recruits nobody — so a mid-upload failure never leaves you with a partially
-configured, live audience.
+Once all your qualification examples are added **and reviewed**, call
+`start_recruiting()` to begin the distilling/onboarding campaign that qualifies
+labelers against them. Recruiting starts only when you call this — until then,
+the audience stays in its `Created` state and recruits nobody, so a mid-upload
+failure never leaves you with a partially configured, live audience.
 
 ```py
 audience.start_recruiting()
 ```
 
 !!! warning "Recruiting is explicit"
-    Earlier SDK versions started recruiting automatically on the first
-    qualification example. That is no longer the case: you must call
-    `start_recruiting()` yourself, after the examples are reviewed. Calling it
-    more than once on the same audience is a no-op.
+    Adding qualification examples does not start recruiting; it is a deliberate
+    step you trigger yourself with `start_recruiting()` once the examples are
+    reviewed. Calling it more than once on the same audience is a no-op.
 
 ### Step 4: Create and Assign a Job
 
--- a/docs/audiences.md
+++ b/docs/audiences.md
@@ -77,9 +77,28 @@ for prompt, datapoint in zip(PROMPTS, DATAPOINTS):
 !!! note
     In practice you'd want to add more examples to the audience to improve the quality of the results.
 
-### Step 3: Create and Assign a Job
+### Step 3: Start Recruiting
 
-Once your audience is set up, create a job definition and assign it to the audience:
+Adding examples no longer starts recruiting on its own. Once all your
+qualification examples are added **and reviewed**, call `start_recruiting()`
+to begin the distilling/onboarding campaign that qualifies labelers against
+them. Until you call it, the audience stays in its `Created` state and
+recruits nobody — so a mid-upload failure never leaves you with a partially
+configured, live audience.
+
+```py
+audience.start_recruiting()
+```
+
+!!! warning "Recruiting is explicit"
+    Earlier SDK versions started recruiting automatically on the first
+    qualification example. That is no longer the case: you must call
+    `start_recruiting()` yourself, after the examples are reviewed. Calling it
+    more than once on the same audience is a no-op.
+
+### Step 4: Create and Assign a Job
+
+Once recruiting has started, create a job definition and assign it to the audience:
 
 ```py
 job_definition = client.job.create_compare_job_definition(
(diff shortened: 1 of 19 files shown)


docs/examples/classify_job.md
```diff
@@ -87,6 +87,8 @@ In this example, we rate images on a Likert scale to assess how well generated i
             settings=[NoShuffleSetting()] # (2)!
         )
 
+    audience.start_recruiting()  # begin recruiting once examples are added and reviewed
```
LinoGiger: please everywhere make sure that this is also with the inline comment of the (3)! for example - so that you can open and close it and it doesn't clutter the stuff

docs/audiences.md
```diff
@@ -77,9 +77,28 @@ for prompt, datapoint in zip(PROMPTS, DATAPOINTS):
...
-### Step 3: Create and Assign a Job
+### Step 3: Start Recruiting
 
-Once your audience is set up, create a job definition and assign it to the audience:
+Adding examples no longer starts recruiting on its own. Once all your
+qualification examples are added **and reviewed**, call `start_recruiting()`
+to begin the distilling/onboarding campaign that qualifies labelers against
+them. Until you call it, the audience stays in its `Created` state and
+recruits nobody — so a mid-upload failure never leaves you with a partially
+configured, live audience.
```
LinoGiger: don't write this as if someone had already read the old documentation. write it just how it is now

docs/audiences.md
```diff
@@ -77,9 +77,28 @@ for prompt, datapoint in zip(PROMPTS, DATAPOINTS):
...
+
+```py
+audience.start_recruiting()
+```
+
+!!! warning "Recruiting is explicit"
+    Earlier SDK versions started recruiting automatically on the first
+    qualification example. That is no longer the case: you must call
+    `start_recruiting()` yourself, after the examples are reviewed. Calling it
+    more than once on the same audience is a no-op.
```
LinoGiger: same here

LinoGiger (changes requested):

docs/examples/classify_job.md
RapidPoseidon: Done — converted the start_recruiting() inline comments to Material collapsible annotations (`# (N)!`) with a footnote, everywhere they appear (classify/compare/draw/locate/select_words job docs, plus confidence_stopping.md, human_prompting.md, and the audiences.md complete example). No more long inline comments cluttering the code.

docs/audiences.md
RapidPoseidon: Done — rewrote this to describe the current behavior directly, without referencing the old auto-recruit-on-first-example behavior.

docs/audiences.md
RapidPoseidon: Done — reworded the admonition to state the current behavior on its own terms (adding examples doesn't start recruiting; call start_recruiting() yourself), without the 'earlier SDK versions...' framing.

LinoGiger (approved):

PR #716, opened 2026-07-24
feat(benchmark): support structured prompt tags (value+category) and origin

RapidPoseidon:
## What

Exposes structured **prompt tags** and a prompt **`origin`** in the Python SDK, plus **editing** and **server-side filter/sort** of prompts.

**SDK leg** of a 3-part full-stack feature:
- Backend: RapidataAI/rapidata-backend#4879 (merged)
- Frontend: RapidataAI/app-frontend#2834

## Contract

- **Tags**: `Tag { value, category? }`. The high-level API accepts a `str | Tag` union (bare strings are wrapped to `Tag(value, category=None)` internally); the wire carries clean `Tag[]` objects.
- **Origin**: additive `Origin { source }`.
- **Update**: prompt tags/origin are editable via `update_prompt(...)`.
- **Filter/sort**: prompts can be queried server-side — filter by `tag_values` / `tag_categories` / `origin_source`; sort by `id` / `created_at` / `origin_source` (tag value/category are filter-only).

## Rebased onto main — the generated client is no longer carried here

This PR previously bundled a **hand-run partial regeneration** of the leaderboard client, because the feature needed `Tag` / `Origin` and the prompt filter/sort params before the spec landed on `main`.

`main` now ships the official generation for this spec (`2026.07.28.0731-56fd9ec`, #725), so that copy has been **dropped entirely** and the SDK binds to the generated models instead.

Verified before dropping it: diffing the branch's generated client against `main`'s, ignoring the OpenAPI-document-version comment line, leaves **only 4 differing files — none related to this feature** (two identity-service fields, a flow TTL description, a version string), all main-side spec updates. The hand-regen was byte-equivalent to what `main` generates, and `main` independently provides everything this feature calls: `sort` / `tag_values` / `tag_categories` / `origin_source` on `benchmark_benchmark_id_prompts_get`, `prompt_api.benchmark_prompt_prompt_id_tags_put`, and `origin` on both prompt input models.

**Net effect: 11 files, +451/−63, with zero churn under `src/rapidata/api_client/` or `openapi/`** — down from 38 files / +1537 spanning ~890 generated files. The `grep -ri tagging src/` → 0 property still holds: the `prompt_tagging.py` / `prompt_origin.py` placeholders are gone from `main`, replaced by generated `models/tag.py` and `models/origin.py`.

Three conflicts resolved while rebasing, worth a reviewer's eye:
- **`main`'s string-tag adaptation** (in #725) touched the same two files. It is **subsumed** by the structured mapping here — the values-only `.tags` view behaves identically, verified at runtime.
- **`uv.lock`** — the pre-rebase branch had dropped every `python_full_version` marker (relocked against a single 3.10 interpreter, losing multi-version resolution). `main`'s lockfile is kept; only the project version field changed, so the lockfile diff is one line.
- **`api_client_README.md`** — generated artifact, took `main`'s.

## Backwards compatibility of the tag input

`tags` is typed as a **covariant `Sequence`**, not `list`, and this matters: `list` is invariant, so typing it as `list[list[str | Tag]]` meant a caller holding `tags: list[list[str]]` — or building one with a comprehension — **no longer type-checked**, even though the value worked fine at runtime. Inline literals slipped through via bidirectional inference, which is what hid it.

All of these are assignable and pass `pyright` today, so pre-category scripts keep checking clean:

```python
legacy:     list[list[str]]        = [["a", "b"]]          # the pre-category shape
structured: list[list[Tag]]        = [[Tag("a", "cat")]]
mixed:      list[list[str | Tag]]  = [["a", Tag("b", "c")]]
with_none:  list[list[str] | None] = [["a"], None]
tuples                             = (("a", "b"),)          # any Sequence
dynamic                            = [t.split(",") for t in ("a,b", "c")]
origins:    list[str]              = ["coco"]               # str, not just Origin
```

`str -> Tag` and `str -> Origin` conversion happens inside the SDK (`coerce_tags` / `coerce_origin` in `prompt_metadata`, shared by `add_prompts` and `update_prompt`), so `category` is genuinely optional and callers never have to construct a `Tag` unless they want one.

One guard worth knowing about: a bare `str` *is* a `Sequence`, so `tags=["a", "b"]` would otherwise be silently read as two prompts' tags instead of one prompt's. `is_tag_list` rejects `str`/`bytes` explicitly so that stays a `ValueError`.

## Changes

**High-level surface**
- User-facing `Tag(value, category=None)`, `Origin(source)`, and `BenchmarkPromptInfo` dataclasses, exported from `rapidata`, `rapidata.rapidata_client`, and `rapidata.types`. These deliberately keep callers off the generated pydantic models; the uploader/mapper translate at the boundary.
- `add_prompts` / `create_new_benchmark`: `tags` accepts `str | Tag` (mixed) per prompt; `origins` accepts `Origin | str`. Bare values are wrapped internally; the uploader sends clean `Tag[]` + `origin`.
- `benchmark.update_prompt(identifier, tags=..., origin=...)` — edits an existing prompt (only the fields passed are changed).
- `benchmark.query_prompts(tag_values=..., tag_categories=..., origin_source=..., sort=...)` — server-side filtered/sorted prompt list, returned as `BenchmarkPromptInfo`.
- Read-back accessors: `.tags` (`list[list[str]]`, values, back-compat), `.structured_tags` (`list[list[Tag]]`), `.origins` (`list[Origin | None]`).

**Version & docs**: `3.17.2` → `3.19.0`. `docs/mri_advanced.md` updated (tags `str | Tag`, `origin`, `update_prompt`, `query_prompts`).

Also folded in while rebasing: the two prompt-mapping helpers were untyped (`prompt` implicitly `Any`) and are now annotated against the generated `GetPromptsByBenchmarkEndpointOutput` under `TYPE_CHECKING`, per the repo's annotation convention.

## Changelog

Prompt tags are now structured — `Tag { value, category? }` — and accepted as a `str | Tag` union (plain strings still work, auto-wrapped). Prompts carry an additive `origin` (`Origin { source }`). New: `benchmark.update_prompt(...)` to edit a prompt's tags/origin, and `benchmark.query_prompts(...)` for server-side filter (tag value/category/origin) and sort (id/created_at/origin_source). Read categories via `benchmark.structured_tags` and origins via `benchmark.origins`; `benchmark.tags` remains the values-only view.

## Validation (re-run post-rebase)

- New `tests/rapidata_client/benchmark/test_prompt_tags.py` (16 cases): tag/origin input shapes incl. legacy `list[list[str]]` and tuples, the values-only `.tags` view, `update_prompt` replacement + cache, and malformed-input rejection.

- `pyright src/rapidata/rapidata_client` → **0 errors** (CI `type-check` green)
- `pytest` → 51 passed, **4 failed — all pre-existing on `main`**, verified by running the suite at the base commit (unrelated audience/job cost-warning tests)
- `uv run --group docs mkdocs build` → succeeds (remaining warnings pre-date this branch)
- `black` clean on every file touched. Two files still fail a repo-wide `black` check — `src/rapidata/__init__.py` and `types/__init__.py` — but that drift **pre-exists on `main`** and is the deliberately-unindented `# --- GENERATED SETTINGS ---` sentinels a codegen script keys on, so they're left alone.
- Runtime checks against `main`'s generated models: create/update wire payloads (`{'tags': [{'value': 'landscape', 'category': 'scene'}], 'origin': {'source': 'coco'}}`), `Tag`/`Origin` read round-trip incl. category and the no-tags/no-origin case, bare-string→`Tag`/`Origin` wrapping, sort-field rejection, `update_prompt` unknown-identifier and empty-update rejection.

🔗 Session: https://poseidon.rapidata.internal/chat/node-ae73dd05
🔗 Rebase session: https://poseidon.rapidata.internal/chat/session-00501664


docs/mri_advanced.md
```diff
@@ -80,8 +106,31 @@ benchmark.add_prompts(
...
+
+```python
+prompts = benchmark.query_prompts(
+    tag_values=["outdoor"],
+    tag_categories=["scene"],
+    origin_source=["coco"],
+    sort=["-created_at"],
 )
+for p in prompts:
+    print(p.identifier, [t.value for t in p.tags], p.origin.source if p.origin else None)
```
LinoGiger: please remove this and related methods

LinoGiger (changes requested):

docs/mri_advanced.md
RapidPoseidon: Thanks @LinoGiger. `update_prompt` and `query_prompts` (plus the `BenchmarkPromptInfo` return type) were added to mirror the cross-service contract for this feature: backend #4879 ships the prompt-update endpoint and the `tag_values`/`tag_categories`/`origin_source` filters + `id`/`created_at`/`origin_source` sort, and frontend #2834 consumes them. I'm glad to drop them from the SDK surface (keeping the core structured tags + origin on create/read/`add_prompts`) — just confirming with the feature owner first, since removing them would diverge the SDK from the other two legs. Will push the removal as soon as that's confirmed.

LinoGiger: just remove the "query_prompts" part with it's methods. it's not needed on the sdk and the interface is clunky. you still have the properties which take care of actual lookups

LinoGiger (changes requested):

docs/mri_advanced.md
RapidPoseidon: Done in 524d2dd — removed `query_prompts` and dropped `BenchmarkPromptInfo` from the public exports (it stays internal for the read path). Kept `update_prompt` per your note, and the existing prompt properties still cover lookups. Ready for another look 🙏

LinoGiger (approved):

PR #746, opened 2026-07-31
feat(context): always shorten over-long contexts, add opt-in contextShortening

RapidPoseidon:
## What

Two separate behaviours, instead of one flag gating both:

**1. Over-long contexts are always shortened.** A context above the backend's 400-character limit is shortened for the job/order instruction before upload, with no way to turn it off — leaving it long only means the backend rejects the datapoint. Because this puts text in front of labelers that the caller did not write, every such rewrite is logged at **warning** level: one summary warning per creation (how many contexts were over the limit) plus a per-datapoint before/after length.

**2. New `rapidata_config.upload.contextShortening`, default `False`.** When enabled, *every* context is shortened for the instruction, not just the over-long ones — a context tuned to the question focuses the labeler even when it already fits. These opted-in rewrites log at info level; the mandatory over-limit ones still warn.

If no instruction is available to shorten against, nothing is shortened and a warning explains why (an over-long context then still gets the "the backend would reject this" warning per datapoint).

## Breaking

`autoShortenContext` is **removed** and replaced by `contextShortening`, whose meaning is different (all contexts vs. only over-long ones), so keeping the old name as an alias would be misleading. It shipped in v3.15.0, so anyone setting `rapidata_config.upload.autoShortenContext` — or the `RAPIDATA_autoShortenContext` env var — needs to drop it: the attribute assignment now raises, and the env var is ignored. Over-long contexts are shortened for them either way, which is what the flag was for. Happy to add a deprecated pass-through instead if you'd rather not break it.

## Notes

- `tests/rapidata_client/context/test_context_manager.py` covers: default-off, mandatory over-limit shortening, no-op when contexts fit, shortening-everything when enabled, the empty-result fallback, and both no-instruction paths.
- Docs updated: `docs/job_definition_parameters.md#contexts` and the upload-config table in `docs/config.md`.
- `pyright src/rapidata/rapidata_client` clean, `black` clean, `mkdocs build` clean. The 4 failing tests in `tests/rapidata_client/audience/` fail identically on `main` — unrelated to this change.

🔗 Session: https://poseidon.rapidata.internal/chat/session-2a59ebc1

### DIFF
--- a/src/rapidata/rapidata_client/context/context_manager.py
+++ b/src/rapidata/rapidata_client/context/context_manager.py
@@ -79,9 +79,10 @@ class ContextManager:
 
         For every datapoint whose context exceeds :data:`MAX_CONTEXT_LENGTH`:
 
-        - if ``rapidata_config.upload.autoShortenContext`` is set and a
-          ``question`` is available, the context is shortened for that question
-          (one batched request) and substituted;
+        - if ``rapidata_config.upload.autoShortenContext`` is set (the default)
+          and a ``question`` is available, the context is shortened for that
+          question (one batched request) and substituted, and a warning reports
+          that it happened;
         - otherwise a warning is logged explaining the backend would reject it.
         """
         over_limit = [
@@ -95,17 +96,15 @@ class ContextManager:
 
         auto_shorten = rapidata_config.upload.autoShortenContext
 
-        if auto_shorten and not question:
-            # Shortening needs the question to tune the context against; without
-            # it we can't shorten, so fall back to warning instead of proceeding.
-            logger.warning(
-                "rapidata_config.upload.autoShortenContext is set but no "
-                "question/instruction was available to shorten against; leaving "
-                "%d over-long context(s) unchanged.",
-                len(over_limit),
-            )
-
         if auto_shorten and question:
+            logger.warning(
+                "%d context(s) exceed the maximum of %d characters and are being "
+                "shortened automatically for the instruction. Set "
+                "rapidata_config.upload.autoShortenContext = False to keep them "
+                "unchanged instead.",
+                len(over_limit),
+                MAX_CONTEXT_LENGTH,
+            )
             shortened = self.shorten_contexts(
                 [(context, question) for _, _, context in over_limit]
             )
@@ -117,7 +116,7 @@ class ContextManager:
                         index,
                     )
                     continue
-                logger.info(
+                logger.warning(
                     "Datapoint %d: shortened context from %d to %d characters.",
                     index,
                     len(context),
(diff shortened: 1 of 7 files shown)


src/rapidata/rapidata_client/context/context_manager.py
```diff
@@ -72,67 +72,85 @@ def shorten_contexts(self, pairs: Sequence[tuple[str, str]]) -> list[str]:
...
+                )
+            if shorten_all and len(over_limit) < len(candidates):
+                logger.warning(
+                    "rapidata_config.upload.contextShortening is enabled but no "
+                    "question/instruction was available to shorten against; leaving "
+                    "%d context(s) unchanged.",
+                    len(candidates) - len(over_limit),
                 )
-                datapoint.context = new_context
             return
```
LinoGiger: why is this here? it's checked everywhere before that this is a string?

LinoGiger (changes requested):
atm we do not parallize the context shortening or display a progress bar. this can feel very unintutive. can you please add both?

src/rapidata/rapidata_client/context/context_manager.py
RapidPoseidon: Removed — you're right, every workflow's `_get_instruction()` returns `str`, so that branch was unreachable. `question` is now typed `str` and the fallback is gone in e72a8c26.

RapidPoseidon:
Both added in `perf(context): shorten contexts in concurrent batches with a progress bar`.

**Parallel + progress:** `shorten_contexts` now splits the pairs into batches of 10 and dispatches them across `rapidata_config.upload.maxWorkers` threads, with a tqdm bar (`Shortening contexts`, respecting `logging.silent_mode`) updating as each batch lands — same shape as the asset-upload and benchmark-prompt paths. Results are reassembled in input order, and a call of <= 10 pairs still goes out as a single request with no bar, so `shorten_context()` for one context is unchanged. Verified with a stubbed endpoint: 25 contexts (3 batches x 0.4s) complete in 0.42s instead of 1.2s.

Worth knowing for the batch size: the endpoint already shortens the items *within* one request concurrently (`Task.WhenAll` in `ShortenContextCommandHandler`) and caches per `(context, question, promptVersion)`. So the win from client-side batching isn't server concurrency — it's that one 1000-context call no longer rides on a single long request with no feedback, and one slow model call can't stall everything behind it. 10 felt like the right balance between request overhead and progress granularity; easy to change if you'd rather it were larger or configurable.

**Log noise:** with a progress bar in place, the per-datapoint before/after line moved back to info. The over-limit warning is now one summary line per creation (`N context(s) exceed the maximum of 400 characters and are being shortened ...`) — a job with a thousand over-long contexts shouldn't emit a thousand warnings, and the mandatory rewrite is still impossible to miss.

LinoGiger (approved):

PR #856, opened 2026-09-03
feat(benchmark): surface min-assets-per-prompt warning on submit

RapidPoseidon:
## What

Surfaces the benchmark **min-assets-per-prompt** warning (added backend-side in [rapidata-backend#5263](https://github.com/RapidataAI/rapidata-backend/pull/5263)) in the Python SDK. When a benchmark sets a minimum number of samples per prompt, submitting a participant now logs a warning naming each prompt it filled with **1..N-1** samples. Submission still succeeds — the check is advisory, never a rejection.

## Changes

1. **`BenchmarkParticipant.run()`** — submits via the **batch endpoint** (`participants_submit_post`) as a batch of one, and logs the warning for the participant. For a single participant this is equivalent to the per-participant submit.
2. **`RapidataBenchmark.run()`** — aggregates the warning across its submit batches and logs one message listing each under-filled participant (by name) and its short prompts.
3. Both use the existing SDK `logger.warning` channel; message shape: `model-a … 'cat' (2/4) …`.

## Why `run()` switched to the batch endpoint

The **single**-submit endpoint's response can't carry the warning yet: its generated schema (`SubmitParticipantEndpointOutput`) collides by name with the *deprecated* `Leaderboard/SubmitParticipantEndpoint`, so the regenerated client only has `validDatapoints`/`invalidDatapoints`. The **batch** endpoint's schema is unique and does carry it. Routing single submits through batch-of-one is the interim path; a small backend rename would let the single endpoint expose the warning and this could revert. (Flagged separately — not in scope here.)

## Testing

- `pyright src/rapidata/rapidata_client` — **0 errors**.
- `black` — clean.
- New tests: `test_participant_run.py` (batch-of-one submit + warning log) and a warning-path case in `test_benchmark_batch_submit.py`. Benchmark suite: **7 passed**.
- Note: 4 pre-existing failures in `tests/rapidata_client/audience/` are unrelated to this change (reproduced on clean `main` before my edits).

🔗 Session: https://poseidon.rapidata.internal/chat/session-f23a2649

### DIFF
--- a/src/rapidata/rapidata_client/benchmark/participant/participant.py
+++ b/src/rapidata/rapidata_client/benchmark/participant/participant.py
@@ -108,11 +108,46 @@ class BenchmarkParticipant:
 
         After uploading media, call this method to submit the participant
         so that it enters the evaluation pipeline.
+
+        If the benchmark requires a minimum number of samples per prompt, any
+        prompt this participant filled with fewer than that many samples is
+        logged as a warning. The submission still goes through — the check is
+        advisory, not a rejection.
         """
-        self._openapi_service.leaderboard.participant_api.participants_participant_id_submit_post(
-            participant_id=self.id
+        from rapidata.api_client.models.submit_participants_endpoint_input import (
+            SubmitParticipantsEndpointInput,
         )
-        self._status = ParticipantStatus.SUBMITTED
+
+        with tracer.start_as_current_span("BenchmarkParticipant.run"):
+            # Submitted through the batch endpoint as a batch of one: only its response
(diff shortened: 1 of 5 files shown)


RapidPoseidon:
Added `RapidataBenchmark.update(min_assets_per_prompt=...)` (commit f48d399) so the gate can be set on an existing benchmark, not just observed at submit. Patches via the update endpoint; validates int ≥ 2 (rejects bool); omitting the arg leaves the stored value unchanged. Covered by `test_benchmark_update.py`.

LinoGiger (approved):

PR #861, opened 2026-09-09
feat(benchmark): set a participant's price for the score-vs-cost chart

RapidPoseidon:
## What

Lets SDK users set a benchmark participant's price so it appears on the benchmark's "Score vs. cost" chart. The backend already accepts `cost` + `costUnit` on the participant create and PATCH endpoints; this PR only adds the SDK surface.

- `BenchmarkParticipant.set_price(price: float, unit: Literal["image", "video_second", "million_tokens"]) -> None` — PATCHes `cost` + `costUnit` only, updates the cached attributes.
- `BenchmarkParticipant.clear_price() -> None` — PATCHes both fields as explicit nulls (the generated pydantic model only serialises a null for fields that were assigned; omitted means "unchanged").
- `participant.price: float | None` and `participant.price_unit: Literal[...] | None`, populated from `benchmark.participants`. Constructor stays backwards compatible (new keyword params default to `None`).
- `add_model` / `evaluate_model` are unchanged. Pricing is participant-level metadata set after the participant exists, per review: adding a model and describing it are separate concerns.
- One private helper (`benchmark/participant/_pricing.py`) owns the unit mapping (`image → UsdPerImage`, `video_second → UsdPerVideoSecond`, `million_tokens → UsdPerMillionTokens`) and validation (finite number > 0, `bool` rejected, unit must be one of the three).

No client regeneration was needed: the list / get / patch models on `main` already carry `cost` and `costUnit`.

## Docs

- `docs/mri.md`: new "Update participant metadata" subsection between 3b and 3c — `rename`, then a Price part: what the price is for, the three units, `set_price` / `clear_price` / `price` read-back example, note that unpriced models are hidden from the chart.
- `docs/mri_advanced.md`: participants listing shows `price` / `price_unit`.
- `CLAUDE.md`: capability mentioned in the list.

## Tests

`tests/rapidata_client/benchmark/test_participant_price.py` (18 cases): unit mapping for all three units, validation errors for price and unit, `clear_price` sends `{"cost": null, "costUnit": null}`, `participants` reads cost back.

## Verification

- `uv run pytest tests` — the 18 new tests pass. 4 pre-existing failures in `tests/rapidata_client/audience` fail identically on `main`.
- `pyright src/rapidata/rapidata_client` — clean after rebasing onto `main` with #864; the `type-check` job is green.
- `black` run on the touched files only; it wanted to reformat 14 unrelated files on `main`, which were left alone to keep this PR focused.

## Out of scope

No faucet changes, no backend changes, no automatic price lookup. The price is whatever the caller passes; "USD per million tokens" is documented as the unit without prescribing how to blend input/output token prices.

## Related

- Docs PR in the skills repo documenting this contract: https://github.com/RapidataAI/skills/pull/10

🔗 Session: [node-d03b9ff7](https://poseidon.rapidata.internal/chat/node-d03b9ff7)

### DIFF
--- a/docs/mri.md
+++ b/docs/mri.md
@@ -152,6 +152,43 @@ participant.run()
 benchmark.run()
 ```
 
+### Update participant metadata
+
+A participant's metadata can be changed after it has been added, independently
+of uploading media or submitting it for evaluation. Any participant works —
+one returned by `add_model` or one taken from `benchmark.participants`.
+
+```python
+participant = benchmark.participants[0]
+
+participant.rename("MyAIModel_v2.2")
+```
+
+#### Price
+
+A model's price places it on the benchmark's "Score vs. cost" chart. The chart
+only shows priced models, and only those quoted in the unit most models in the
+benchmark use — unpriced models stay on the leaderboard but are hidden from the
+chart.
+
+Prices are in USD per one of three units:
+
+| `unit`             | Meaning                       |
+| ------------------ | ----------------------------- |
+| `"image"`          | USD per generated image       |
+| `"video_second"`   | USD per second of video       |
+| `"million_tokens"` | USD per million tokens        |
+
+```python
+participant.set_price(0.04, "image")
+
+print(participant.price, participant.price_unit)  # 0.04 image
+
+participant.clear_price()  # hides the model from the chart again
+```
+
+`price` and `price_unit` are `None` for models without a price.
+
 ### 3c. Recovering a Partial Upload
 
 Large uploads over a slow or flaky connection can leave individual samples
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -7,7 +7,7 @@ As a customer you can use the RapidataClient class to access the following:
 - AudienceCreation (including filtered audiences via `.filter()` for country/language/demographic targeting)
 - ValidationSetCreation
 - FlowCreation
-- BenchmarkCreation / MRI Creation
+- BenchmarkCreation / MRI Creation (including participant metadata such as `participant.rename()` and `participant.set_price()` for the score-vs-cost chart)
 
 Orders were removed from the SDK entirely (v3.21.0) — do not add, document, or reference order creation.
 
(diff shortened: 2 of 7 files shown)


RapidPoseidon:
CI note: the `type-check` failure is not from this change. The 4 pyright errors (`_prompt_uploader.py:65-66`, `rapidata_benchmark.py` `prompt_asset`) are the same 4 that fail on `main` since the OpenAPI client regeneration in #857 (`pyright-type-checking` has been red on every `openapi-generation-main` run since 2026-09-08). They live in the prompt-asset code, which this PR does not touch; #859 addresses them. Once #859 lands I'll rebase this branch and re-run CI.

Verified locally: `pyright src/rapidata/rapidata_client` reports the identical 4 errors on a pristine `main` checkout.

RapidPoseidon:
Rebased onto current `main` (includes #864 and the #862 client regeneration). `pyright src/rapidata/rapidata_client` is now clean and the `type-check` job passes on 2bcf36c0. The 25 pricing tests pass; the 4 remaining failures in `tests/rapidata_client/audience` fail identically on `main` and are unrelated.

docs/mri.md
```diff
@@ -152,6 +152,42 @@ participant.run()
...
+| `"million_tokens"` | USD per million tokens        |
+
+Set the price when adding the model:
+
+```python
+participant = benchmark.add_model(
+    name="MyAIModel_v2.1",
+    media=media,
+    prompts=prompts,
+    price=0.04,
```
EliasTeikari: lets change the interface for this, adding a participant to the benchmark and adding the price for the participant should be separate concerns, so we shouldnt make that into one call. it should be an extra section like "update participant metadata"

EliasTeikari (changes requested):

docs/mri.md
RapidPoseidon: Done in f6551767. `add_model` / `evaluate_model` no longer take `price` / `price_unit`; the create payload is back to `name` only. Pricing is now participant-level only: `participant.set_price(price, unit)`, `participant.clear_price()`, and the `price` / `price_unit` read-back. This section became "Update participant metadata", grouping `rename` with the price operations; the add_model pricing example is gone from the docs and CLAUDE.md.

EliasTeikari (approved):
lgtm

PR #881, opened 2026-09-19
docs(flow): dogfood classify flow docs on 3.24.1

RapidPoseidon:
## STATUS: finished, one thing still blocked

The backend shipped (`2026.09.19.2045-5a20f08`, then `2026.09.19.2051-b65081c` after `responsesRequired` was removed entirely). The client is regenerated (reconciled with `main`'s own auto-regeneration rather than carrying a duplicate one -- see "Amendment history" #5) and `create_classify_flow` now sends real `maxResponses`/`minResponses`. `max_datapoints_per_item` is fully removed.

**One thing is still open**: whether to drop the SDK's client-side zero-fill of `distribution` now that the results endpoint is documented to return every blueprint category itself. I could not verify this empirically -- `GET /flow/simple/item/{flowItemId}/results` is currently returning a 400 (`'Size' must be less than or equal to '100'`) for every classify item, old and brand new, confirmed as a real backend regression with a hotfix in progress. Per "keep the fill if in doubt," the zero-fill stays for now. Will remove it once told the hotfix is deployed and I can verify against the named production item.

## What this is

A from-scratch dogfood of the classify flow feature (rapidata 3.24.1, docs republished 2026-09-19 18:56 UTC) against production, as a brand-new customer would experience it: fresh venv, docs only, then the SDK against real credentials. Full method: create a classify flow, both `create_new_flow_batch` shapes (plain and with `contexts=`), block on `get_results()`, inspect every result field, `get_flow_by_id`, `find_flows`, `get_flow_items`, `update_config`, and every documented validation/error path (category count bounds, duplicate category values, TTL bounds, the `max_datapoints_per_item` cap).

**This PR has since gone through three rounds of follow-up requirements from the requester** — see "Amendment history" near the bottom, which explicitly flags two places where a later round reverses earlier flow-level TTL work.

## Mismatches found and fixed (current state)

| # | Claim (docs/flows.md, before this PR) | Observed | Fix |
|---|---|---|---|
| 1 | Line 196: "Every response is billed at your organization's per-response rate. A batch collects `responses_per_datapoint` responses for each of its items, so a full batch with the defaults is billed as 5 × 24 = 120 responses." | Per direction from the requester, the org-rate framing should not appear in the public docs. | Reworded to "Each response is billed. A batch collects `responses_per_datapoint` responses for each of its items, so a full batch with the defaults comes to 5 × 24 = 120 responses." |
| 2 | The classify section walked through a second, "advanced" `create_classify_flow` example: `(label, value)` tuple categories, `responses_per_datapoint`, `max_datapoints_per_item`, `time_to_live`, with three numbered footnotes. | Per a later request from the requester: keep only the plain string-categories example in the guide. Tuples stay fully supported in the SDK and documented in the docstring/reference — they're just not walked through here. | Removed the advanced example and its footnotes. The defaults they used to explain are now one short sentence of prose, so the billing-math sentence right below it (row 1) still has the numbers it refers to. |
| 3 | The advanced example (row 2) showed a flow-level `time_to_live` parameter on `create_classify_flow`. | Per the requester: classify flows should work exactly like ranking flows for TTL — configurable only per batch, never at flow creation. (This went through an intermediate step first; see "Amendment history.") | Removed `time_to_live` from `create_classify_flow` entirely: signature, docstring, timedelta handling, bounds check, and the `defaultTimeToLiveSeconds` wire field. The backend's 4-minute default applies automatically. The only TTL a customer sets is `create_new_flow_batch(time_to_live: int \| None)`, 45s–3600s — documented with the same wording the ranking-flow guide already uses. |
| 4 | `ClassifyDatapointResult.distribution` (docs and docstring) said categories nobody chose are omitted from the mapping. | Per the requester: results should carry every category of the flow's blueprint, in blueprint order, with `0` for categories nobody chose. A backend fix for this is in progress but not shipped yet. | `RapidataFlowItem._get_classify_results` now fetches the flow once via `flow_api.flow_flow_id_get` and merges the full ordered blueprint category list into each datapoint's distribution client-side (0-filling; any category the API returns that isn't in the blueprint is appended after). Docs and docstring updated to match. Added `test_distribution_includes_every_blueprint_category`. |
| 5 | The guide explained `max_datapoints_per_item` as a tunable parameter: name, default (24), upper bound (100). | Per the requester: the backend is retiring this parameter (moving to a fixed 256-per-item limit everyone gets). It still works today, but shouldn't be taught as a knob going forward. The parameter itself stays in the SDK for now — removal is a follow-up once the backend ships. | Removed it from the guide's prose entirely. Left the parameter and its validation untouched in code; added a one-line deprecation note to its docstring. |

No other mismatches survived verification. Everything below was run against production during the initial dogfood pass and matched the docs/docstrings/reference exactly:

- `create_classify_flow`'s category-count bound (2–10), rejected 0/1/11 categories client-side with `ValueError: Categories must contain between 2 and 10 entries.`
- Duplicate category values rejected client-side for both plain-string and `(label, value)` tuple forms: `ValueError: Category values must be unique.`
- `time_to_live` bounds (45s–3600s) rejected client-side at both the flow level and the batch level with `ValueError: Time to live must be between 45 seconds and 1 hour.`
- `max_datapoints_per_item`: defaults to 24, accepts up to 100, rejects `0`/negative client-side (`ValueError: Max datapoints per item must be at least 1.`) and rejects 101 server-side (`RapidataError: 'Max Datapoints Per Item' must be between 1 and 100.`). Submitting 25 datapoints to a flow with the default cap of 24 was rejected server-side with `Dataset must have at most 24 datapoints, but has 25` — exactly as documented ("larger batches are rejected").
- `create_new_flow_batch` with `contexts=` (one string per image) worked as documented.
- `get_results()` blocked until completion and returned a `ClassifyFlowItemResult` keyed by source URL, with `distribution` correctly omitting categories nobody chose, and `total_responses` matching the sum of `response_count` across datapoints.
- `get_response_count()` matched `total_responses`.
- `get_win_loss_matrix()` on a classify flow item raised `ValueError: The win/loss matrix is only available for ranking flow items.`, exactly as documented.
- `get_flow_by_id`, `find_flows`, and `get_flow_items` all round-tripped our flow/items correctly.
- `update_config()` on a classify flow raised `ValueError: update_config is only available for ranking flows.`, exactly as documented.
- `flow.delete()` worked (used for cleanup).

## Not fixed here — flagged for the team

- **`majority_value=None` on a tie, and the `Incomplete` state for a zero-response batch** are both documented claims I could not exercise: our real annotator responses never landed in an exact tie, and every batch we ran got at least one response before we could observe a true TTL-expiry-with-zero-responses. Not contradicted, just unverified.
- **`create_classify_flow` accepts `validation_set_id` and `settings`** (confirmed in source/docstring/reference) but neither is mentioned anywhere in `docs/flows.md`'s classify section. Not a factual error, so left out of this PR to keep it small, but a customer wanting a gold-standard validation set or custom settings on a classify flow has no way to discover that from the guide alone.
- **Site information architecture**: `docs/flows.md`'s "Classify Flows" section and the separate "Examples → Classification" nav entry (`docs/examples/classify_job.md`) document two different, both-currently-supported APIs — the new `client.flow.create_classify_flow` blueprint system vs. the older `client.job.create_classification_job_definition` / `audience.assign_job` Likert-scale system. Neither page links to the other. Flagging as a product/IA call, not a docs bug.

## Min/max responses per datapoint (prep, blocked on backend)

**Contract** (mirrors ranking's own min/max response threshold, per datapoint instead of per item):

- `maxResponses` (per datapoint, default 5, >= minResponses): the referee stop count -- an image closes once it has this many accepted responses.
- `minResponses` (per datapoint, default 3, >= 1): the acceptance floor. If an item ends by `time_to_live`, it's `Completed` if total responses >= `minResponses x datapoint count`, otherwise `Incomplete` (same total-based check ranking already uses). Items that end because every image hit `maxResponses` are `Completed` regardless.
- `responsesRequired` is being renamed to `maxResponses` backend-wide; since SDK 3.24.1 is already out sending `responsesRequired`, the endpoint inputs keep it as an `[Obsolete]`, OpenAPI-deprecated alias mapping to `maxResponses` when `maxResponses` is absent, until a later backend release deletes it.
- `max_datapoints_per_item` is being removed once the regenerated model drops the field (backend fixes it at 256).

**Done now** (commit `81791d4`, doesn't touch the wire format):

- `create_classify_flow` gains `max_responses_per_datapoint` (default 5) and `min_responses_per_datapoint` (default 3), validated exactly like the contract above (`min >= 1`, `max >= min`).
- `responses_per_datapoint` becomes a deprecated, keyword-only alias: passing it sets both max and min to that value and raises a `DeprecationWarning`.
- Docs rewritten to describe the threshold range the way the ranking section describes its own -- bullet list, a second example with footnotes, and a matching Completed/Incomplete note. Billing sentence now says "up to `max_responses_per_datapoint` responses per image" instead of a fixed count.
- Tests: bounds validation for both new parameters, the deprecated alias (value + warning), and the payload shape.

**Deferred to the finishing commit** (needs the regenerated client, which doesn't have `maxResponses`/`minResponses` yet -- only `responsesRequired` and `maxDatapointsPerItem`):

- Sending `maxResponses` / `minResponses` on the wire instead of the interim `responsesRequired=max_responses_per_datapoint` shim. There's a two-line comment on that line in `rapidata_flow_manager.py` marking exactly why.
- Removing `max_datapoints_per_item` from the signature, docstring, validation, and docs once the regenerated model no longer has the field.
- `min_responses_per_datapoint` is fully validated but has **no effect yet** -- there's nowhere to send it until regeneration. This is deliberate, not an oversight.

## Amendment history

This PR went through three rounds of follow-up requirements from the requester after the initial dogfood pass. All were applied as new commits on this same branch, per instructions, rather than opening new PRs.

1. **Commit `56dc3b4`** (initial): fixed the billing sentence, and — based on my own dogfood testing — changed the guide's advanced example from `time_to_live=timedelta(minutes=4)` to `timedelta(minutes=10)`. Reasoning at the time: the real SDK default was `None`, not 4 minutes, so showing an explicit *non-default* override was more honest than a code sample that happened to match the default value while the signature said otherwise.
2. **Commit `71ac9e8`**: **reverses that specific change.** The requester decided the SDK should default to 4 minutes explicitly (`time_to_live: timedelta | int = timedelta(minutes=4)`, rather than `None` plus a backend fallback), which made the original `timedelta(minutes=4)` example correct again, so it was restored. Also independently removed the advanced tuple-categories example from the guide per a second, separate requirement. Commit 1 wasn't wrong when written — it accurately described the SDK *before* commit 2 changed it.
3. **Commit `d62f0b9`** (this one): **reverses commit 2's flow-level TTL work entirely, and does so intentionally at the requester's direction.** Classify flows now work exactly like ranking flows for TTL: `time_to_live` is gone from `create_classify_flow` completely (no parameter, no default, no validation, nothing sent on the wire) — the backend's 4-minute default just applies, and the only TTL a customer ever sets is per batch. This also meant deleting `test_time_to_live_defaults_to_four_minutes` (added in commit 2, now describing a parameter that no longer exists) and the two flow-level TTL-bounds cases from the parametrized rejection test, and fixing two tests that still asserted the old `defaultTimeToLiveSeconds` wire behavior. Commit 2 wasn't wrong when written either — each commit accurately reflected the SDK as it stood at that point; the requester's understanding of what they wanted here evolved twice in a row.

This same commit independently ships two unrelated fixes: `ClassifyDatapointResult.distribution` now includes every category of the flow (not just the ones with votes), and `max_datapoints_per_item` gets a docstring deprecation note ahead of a planned backend removal.

Net effect: the guide shows only the plain example, TTL is documented and configured solely at the batch level (identically to ranking), distributions are always complete, `max_datapoints_per_item` is quietly on notice, and the billing sentence no longer mentions an organization-level rate.

4. **Commit `81791d4`** (this one): **not a reversal** -- new, additive prep work. The requester decided classify flows need a min/max response-per-datapoint range in v1 (not a fixed count), because a fixed-then-ranged rollout later would be a breaking change, and because the PID controller needs the min/max leeway -- with a fixed count every item lands exactly on target, the error pins at zero, and the controller sticks to `pidMin` (confirmed by their load test). See "Min/max responses per datapoint" below for the full contract and what's deferred to the finishing commit.

## Finishing commit (this round)

- `create_classify_flow` sends `maxResponses`/`minResponses` on the wire instead of the interim `responsesRequired` shim -- the shim is gone entirely now that the field is removed from the spec.
- `max_datapoints_per_item` removed completely: signature, docstring, validation, wire field. It was never in the docs (removed from the guide two rounds ago), so no docs change needed there.
- Client-side distribution zero-fill **kept**, not removed -- see the status banner above.
- Docs: billing sentence's worked example now uses the real fixed cap (5 × 256 = 1280) instead of the retired `max_datapoints_per_item` default of 24. Added one sentence to the "Create a Flow" section pointing to where `time_to_live` actually lives now, per @LucStr's review comment (docs/flows.md:192) -- replied on the PR.
- Reconciled with `main`: while this PR was open, `main` picked up two automated OpenAPI client regenerations of its own (`8e4e566b3`, then `989f78635` after the `responsesRequired` removal). Rather than carry this branch's own hand-run regeneration (which was one deploy behind), merged `main` in and took its generated client wholesale -- see "Amendment history" #5.
- Also discovered while investigating: `main` already had #880 merged (an earlier, independent dogfood pass against 3.24.0, 17 mismatches found/15 fixed) before this branch was ever created, so its fixes were already this branch's baseline the whole time -- nothing to reconcile there, just noting it for anyone reading the history.

Tests: replaced `test_max_responses_per_datapoint_is_sent_as_responses_required` with `test_max_and_min_responses_per_datapoint_are_sent` (checks both `maxResponses` and `minResponses` in the payload), updated `test_responses_per_datapoint_alias_is_deprecated` and `test_payload_matches_generated_model` for the new field names, removed the now-invalid `max_datapoints_per_item` parametrize case.

## Small correction (latest commit)

Per feedback: dropped the highlighted `!!! note` callout for the TTL forward-reference (back to plain prose, now with a working link to the batch section instead of vague "see the next section"), and raised `max_responses_per_datapoint`/`min_responses_per_datapoint` defaults from 5/3 to 15/10 -- a deliberate SDK-level default, diverging from the backend's own stated default of 5/3. Updated everywhere the old defaults were stated: signature, docstring, docs bullet list, the billing worked example (now 15 × 256 = 3,840), and the one test asserting default payload values.

## Local checks (re-run after every commit, latest numbers below)

- `uv run --group docs mkdocs build` — succeeds (pre-existing unrelated warnings only, e.g. `mri.md`/`box.py`, not touched by this change).
- `uv run pyright src/rapidata/rapidata_client` — 0 errors, 0 warnings.
- `uv run black --check src/rapidata/rapidata_client/flow/ tests/rapidata_client/flow/` — clean. (Repo-wide `black --check` still flags the same 11 pre-existing, unrelated files it did before this PR; nothing in this diff.)
- `uv run pytest tests/rapidata_client/flow/` — 32 passed. `uv run pytest tests/ --ignore=tests/test_docs_site.py` — 224 passed, 4 pre-existing failures in `tests/rapidata_client/audience/` unrelated to flows (confirmed identical with none of this PR's changes applied). `test_docs_site.py` fails to collect (`ModuleNotFoundError: No module named 'scripts'`) regardless of this PR -- pre-existing path issue, not touched here.

## Production run: ids and scripts

From the initial dogfood pass (no new production calls were needed for the two follow-up amendments — those were pure docs/code/test changes). Account: `poseidon@rapidata.ai`. Flow (soft-deleted after the run via `flow.delete()`): `flw_1Vd60zi3gir9gp`. Flow items: `fli_1Vd61olekWb7ak` (plain batch, 2 images), `fli_1Vd62WPhShks32` (batch with `contexts=`, 2 images). Real annotator responses: 10 per item, 20 total, plus a handful more from the error-path probes below.

<details>
<summary>phase1a_create.py — flow creation, both batch shapes, every validation/error path</summary>

```python
import json
import sys
import traceback

from rapidata import RapidataClient

def step(label, fn):
    print(f"\n=== {label} ===", flush=True)
    try:
        result = fn()
        print(f"OK: {result!r}", flush=True)
        return result, None
    except Exception as e:
        print(f"RAISED: {type(e).__name__}: {e}", flush=True)
        return None, f"{type(e).__name__}: {e}"

client = RapidataClient()

step("find_flows sanity check (amount=3)",
     lambda: [(f.id, getattr(f, 'name', None)) for f in client.flow.find_flows(amount=3)])

IMG_1 = "https://assets.rapidata.ai/9839a335-d2e9-45ff-871d-5002c763c4b3.png"
IMG_2 = "https://assets.rapidata.ai/488841b8-9bb3-40a0-83b4-5ba5400e1118.png"
IMG_3 = "https://assets.rapidata.ai/53c22735-ad61-4c4e-8f68-742e9df7b747.webp"
IMG_4 = "https://assets.rapidata.ai/a52bc419-bd92-45fb-aaba-8e5c8b9942f5.webp"

flow, err = step(
    "create_classify_flow (simple form, matches docs verbatim)",
    lambda: client.flow.create_classify_flow(
        name="Text Detection",
        instruction="Does this image contain text?",
        categories=["Yes", "No"],
    ),
)
if flow is None:
    print("FATAL: could not create main flow, aborting")
    sys.exit(1)

print("flow.id =", flow.id)
for attr in ("name", "flow_type", "id"):
    print(f"flow.{attr} =", getattr(flow, attr, "N/A"))

ids = {"flow_id": flow.id, "stray_flow_ids": [], "stray_item_ids": []}

def try_bad_flow(label, **kwargs):
    base = dict(name=f"dogfood-{label}", instruction="test", categories=["A", "B"])
    base.update(kwargs)
    f, _ = step(f"error path: {label}", lambda: client.flow.create_classify_flow(**base))
    if f is not None:
        ids["stray_flow_ids"].append(f.id)
        print(f"  NOTE: flow WAS created despite invalid input! id={f.id}")
    return f

try_bad_flow("zero-categories", categories=[])
try_bad_flow("one-category", categories=["OnlyOne"])
try_bad_flow("eleven-categories", categories=[str(i) for i in range(11)])
try_bad_flow("dup-plain-categories", categories=["Yes", "Yes"])
try_bad_flow("dup-tuple-values", categories=[("Yes, clearly", "yes"), ("Also yes", "yes")])
try_bad_flow("ttl-too-low", time_to_live=10)
try_bad_flow("ttl-too-high", time_to_live=7200)

item1, _ = step(
    "create_new_flow_batch (plain, 2 images)",
    lambda: flow.create_new_flow_batch(datapoints=[IMG_1, IMG_2]),
)
item2, _ = step(
    "create_new_flow_batch (with contexts=, 2 images)",
    lambda: flow.create_new_flow_batch(
        datapoints=[IMG_3, IMG_4],
        contexts=["First submitted face photo.", "Second submitted face photo."],
    ),
)

if item1: ids["item1_id"] = item1.id
if item2: ids["item2_id"] = item2.id
print("item1.id =", getattr(item1, "id", None))
print("item2.id =", getattr(item2, "id", None))

bad_ttl_item, _ = step(
    "error path: batch time_to_live too low (20s)",
    lambda: flow.create_new_flow_batch(datapoints=[IMG_1], time_to_live=20),
)
if bad_ttl_item:
    ids["stray_item_ids"].append(bad_ttl_item.id)
    print("  NOTE: batch WAS created despite TTL=20 (below documented 45s min)! id=", bad_ttl_item.id)

big_item, _ = step(
    "error path: 25 datapoints exceeding default max_datapoints_per_item=24",
    lambda: flow.create_new_flow_batch(datapoints=[IMG_1] * 25),
)
if big_item:
    ids["stray_item_ids"].append(big_item.id)
    print("  NOTE: batch WAS created despite 25 > default max=24! id=", big_item.id)

step("item1.get_status()", lambda: item1.get_status())
step("item2.get_status()", lambda: item2.get_status())

flow_again, _ = step("get_flow_by_id(flow.id)", lambda: client.flow.get_flow_by_id(flow.id))
if flow_again:
    print("flow_again.id == flow.id:", flow_again.id == flow.id)
    print("flow_again.name =", getattr(flow_again, "name", "N/A"))

found, _ = step("find_flows(amount=10)",
                lambda: [(f.id, getattr(f, 'name', None)) for f in client.flow.find_flows(amount=10)])
if found:
    print("our flow.id present in find_flows:", flow.id in [f[0] for f in found])

items, _ = step("flow.get_flow_items(amount=10)", lambda: [i.id for i in flow.get_flow_items(amount=10)])
if items:
    print("item1 present:", getattr(item1, 'id', None) in items)
    print("item2 present:", getattr(item2, 'id', None) in items)

step("flow.update_config(instruction='new instruction')",
     lambda: flow.update_config(instruction="new instruction"))

with open("phase1_ids.json", "w") as fh:
    json.dump(ids, fh, indent=2)
print("\n=== IDS SAVED ===")
print(json.dumps(ids, indent=2))
```

</details>

<details>
<summary>phase1b_wait_results.py — blocks for real annotator responses, inspects results, cleans up</summary>

```python
import json

from rapidata import RapidataClient

with open("phase1_ids.json") as fh:
    ids = json.load(fh)

client = RapidataClient()
flow = client.flow.get_flow_by_id(ids["flow_id"])
items = {i.id: i for i in flow.get_flow_items(amount=10)}
item1 = items[ids["item1_id"]]
item2 = items[ids["item2_id"]]

print("Pre-wait status: item1=", item1.get_status(), "item2=", item2.get_status(), flush=True)

results1 = item1.get_results()
print("\n=== item1 results ===", flush=True)
print("type:", type(results1))
print("total_responses:", getattr(results1, "total_responses", "N/A"))
for k, v in results1.datapoints.items():
    print(f"  key={k!r} majority_value={v.majority_value!r} distribution={v.distribution!r} response_count={v.response_count!r}")

results2 = item2.get_results()
print("\n=== item2 results ===", flush=True)
print("type:", type(results2))
print("total_responses:", getattr(results2, "total_responses", "N/A"))
for k, v in results2.datapoints.items():
    print(f"  key={k!r} majority_value={v.majority_value!r} distribution={v.distribution!r} response_count={v.response_count!r}")

print("\nitem1.get_response_count():", item1.get_response_count(), flush=True)
print("item2.get_response_count():", item2.get_response_count(), flush=True)

try:
    item1.get_win_loss_matrix()
    print("get_win_loss_matrix() on classify item: NO ERROR RAISED (unexpected!)")
except Exception as e:
    print(f"get_win_loss_matrix() on classify item raised: {type(e).__name__}: {e}")

print("\nFinal status: item1=", item1.get_status(), "item2=", item2.get_status(), flush=True)

try:
    flow.delete()
    print("flow.delete() OK")
except Exception as e:
    print(f"flow.delete() raised: {type(e).__name__}: {e}")

print("\n=== DONE ===")
```

</details>

Raw output logs from both runs are in the session transcript if anyone wants the full trace, including the actual response distributions we got back (e.g. `4/1` and `1/4` splits on the two face photos — no ties this run).

Session: [poseidon.rapidata.internal/chat/node-15e5acc8](https://poseidon.rapidata.internal/chat/node-15e5acc8)


RapidPoseidon:
Two follow-up requirements came in from the requester after the initial dogfood commit (`56dc3b4`), both applied here as commit `71ac9e8`:

1. **Keep classify TTL at 4 minutes, but make it the SDK's real default.** This directly reverses part of my first commit: I had changed the guide's example from `time_to_live=timedelta(minutes=4)` to `timedelta(minutes=10)` because the SDK's actual signature default was `None`, not 4 minutes. The requester decided the SDK itself should default to `timedelta(minutes=4)` explicitly instead (same wire behavior — the backend already treated a null `defaultTimeToLiveSeconds` as 240s). That makes the original `timedelta(minutes=4)` example correct again, so it's restored. My first commit wasn't wrong when I wrote it — it accurately described the SDK before this commit changed it.
2. **Remove the advanced tuple-categories example from the guide entirely** (separate, independent request). Tuples are still fully supported in code and the docstring/reference, just not walked through in `docs/flows.md` anymore.

PR body updated to match. CI re-run: pyright clean, black clean on the touched files, all 33 flow tests pass (including a new one asserting the 4-minute default). Watching checks now.

RapidPoseidon:
Third round of follow-up requirements landed as commit `d62f0b9`. One of them is a full reversal of commit `71ac9e8`'s flow-level TTL work, flagging per instructions:

**Reversed:** `create_classify_flow` no longer has a `time_to_live` parameter at all — not `None`, not `timedelta(minutes=4)`, nothing. TTL now works exactly like ranking flows: unconfigurable at flow creation (the backend's 4-minute default just applies), settable only per batch via `create_new_flow_batch(time_to_live: int | None)` in seconds, 45–3600. This undoes both my original dogfood fix (which changed the guide's example to a non-default override because the real default was `None`) and the immediately-preceding amendment (which made `timedelta(minutes=4)` the SDK's actual explicit default). Neither earlier commit was wrong when written — each accurately described the SDK at that point in time. Deleted `test_time_to_live_defaults_to_four_minutes` (added last round, now testing a parameter that no longer exists) and the two flow-level TTL-bounds cases from the parametrized rejection test; fixed two more tests that still asserted the old wire behavior.

**Not a reversal, new work:**
- `ClassifyDatapointResult.distribution` now always lists every category of the flow's blueprint (0-filled where nobody voted), computed client-side in `RapidataFlowItem._get_classify_results` via one extra `flow_api.flow_flow_id_get` call per results fetch, until the backend does this itself. New test: `test_distribution_includes_every_blueprint_category`.
- `max_datapoints_per_item` gets a one-line docstring deprecation note (backend is moving to a fixed 256-per-item limit); removed from the guide's prose; parameter and validation untouched in code.

Re-ran everything: pyright clean, black clean, `mkdocs build` succeeds, all 30 flow tests pass. Watching CI now.

RapidPoseidon:
Commit `81791d4`: prep for min/max responses per datapoint on classify flows, per the requester. **This one does not reverse anything already shipped** -- it's additive.

**Holding this PR as a draft per the requester -- do not merge.** The backend change (classify flows get `maxResponses`/`minResponses`, mirroring ranking's own threshold contract) is landing within hours; this commit prepares the SDK side without depending on it.

What changed now: `create_classify_flow` takes `max_responses_per_datapoint` (default 5) and `min_responses_per_datapoint` (default 3), validated the same way the contract specifies (min >= 1, max >= min). `responses_per_datapoint` is now a deprecated, keyword-only alias (sets both, warns). Docs rewritten to describe the threshold range the way the ranking section already describes its own. Full contract and the done-now/deferred split are in the PR body under "Min/max responses per datapoint."

What's still pending: the actual wire fields. The currently-installed generated client only has `responsesRequired`/`maxDatapointsPerItem` -- no `maxResponses`/`minResponses` yet -- so the SDK still sends `responsesRequired=max_responses_per_datapoint` as an interim shim (commented in the source), and `min_responses_per_datapoint` is validated but has no effect on the wire yet. `max_datapoints_per_item` stays untouched. Finishing this -- swapping the wire fields, dropping `max_datapoints_per_item` -- happens once the requester confirms the backend is deployed and the client is regenerated.

Re-ran everything: pyright clean, black clean, mkdocs build succeeds, 33 flow tests pass (2 new: alias behavior + the new payload shape). Not watching CI further or touching the draft/ready status -- nothing else is expected here until the finishing signal arrives.

docs/flows.md
```diff
@@ -174,28 +174,29 @@ flow = client.flow.create_classify_flow(
...
     instruction="Does this image contain text?",
-    categories=[("Yes, clearly readable", "yes"), ("No", "no")],
-    responses_per_datapoint=5, # (1)!
-    max_datapoints_per_item=24, # (2)!
-    time_to_live=timedelta(minutes=4), # (3)!
+    categories=["Yes", "No"],
+    max_responses_per_datapoint=8, # (1)!
+    min_responses_per_datapoint=4, # (2)!
 )
 ```
```
LucStr, line 194: like time_to_live should be integer and document not entirely removed!

RapidPoseidon:
Re: "time_to_live should be integer and document not entirely removed" (docs/flows.md:192)

Not silently dropped -- it moved. TTL for classify flows now works exactly like ranking flows, per explicit direction from the requester earlier in this PR (see "Amendment history" in the description): it's not a flow-creation-time parameter at all anymore, only a per-batch one via `create_new_flow_batch(time_to_live=...)`. That's documented as a plain integer in the "2. Add a Flow Batch" section right below (`time_to_live=120`), matching your preference for an int over a `timedelta`.

Fair point that dropping it from the "Create a Flow" example without a trace reads like an oversight rather than a deliberate move -- pushed a follow-up that adds one sentence right where it used to be: "`time_to_live` works differently: it isn't set here at all, only per batch -- see the next section." Same commit also finishes wiring `max_responses_per_datapoint`/`min_responses_per_datapoint` to the now-deployed backend (`maxResponses`/`minResponses`) and drops `max_datapoints_per_item` entirely, per the requester's latest message -- full details in the updated PR body.

PR #886, opened 2026-09-20
fix(telemetry): suppress traces from test harnesses that are not pytest

LucStr:
## Problem

In the last 24h, **204 of 645 error traces across all of Rapidata (32%)** came from `Rapidata.Python.SDK` spans whose `StatusMessage` contains `MagicMock` — a test harness driving the SDK with fakes, exporting to the production collector.

```
2026-09-19 17:25  3.24.0   50 errors (43 MagicMock)
2026-09-19 19:30  3.24.1   43 errors (38 MagicMock)
2026-09-19 21:43  3.25.0   49 errors (43 MagicMock)
2026-09-19 23:06  3.25.1   42 errors (37 MagicMock)
```

Over 7 days the split is unambiguous — SDK spans that carry an `identity` attribute (a real `RapidataClient` authenticated) have a **0.016%** error rate; spans without one have **6.5%**. The mock bursts are all in the second group.

`#867` already added a pytest guard, and it works: under pytest the default is `False` and the tracer never initialises. But the bursts above come from **3.25.1, which ships that guard**, so the producer is not a pytest process. Two holes let it through:

1. `_running_under_pytest()` only knows pytest. A harness using `unittest.mock` without pytest — a plain script, a `subprocess` spawned from a test, a notebook — looks like a customer.
2. `_default_enable_otlp()` is evaluated **once**, when `rapidata_config` is constructed at import. A module that does `import rapidata` before `from unittest.mock import MagicMock` has already fixed the default to `True`. `RapidataTracer._enabled` also starts at `True` and only ever changes on a config notification.

Reproducing the production failure mode exactly, on `main`:

```python
import rapidata
from rapidata.rapidata_client.config import tracer
from unittest.mock import MagicMock          # imported after rapidata
with tracer.start_as_current_span("RapidataAudience.add_locate_example"):
    pass
# tracer._enabled = True -> exports to otlp-sdk.rapidata.ai
```

## Change

- `_running_under_pytest()` becomes `_running_under_test()` and also treats an imported `unittest.mock` as a test runner. A process that imported the mock library is driving the SDK with fakes, whatever runs it.
- `RapidataTracer._ensure_initialized()` re-checks the test-runner signal and `RAPIDATA_DISABLE_OTLP` **at first span** rather than trusting the import-time default. By then a harness has imported its mocks, so the late check catches the ordering above.
- The late check is skipped when the user set `enable_otlp` explicitly (tracked via `model_fields_set`), so `LoggingConfig(enable_otlp=True)` and `rapidata_config.logging.enable_otlp = True` still force export on — including from a test, which is what someone debugging SDK telemetry wants.

Customer behaviour is unchanged: a process that never imports a mock library still exports on the first span, verified below.

## Verification

```
uv run python -m pytest tests/rapidata_client/config/test_otlp_default.py -q   ->  10 passed
uv run python -m pytest -q                                                    ->  229 passed, 4 failed
uv run pyright src/rapidata/rapidata_client                                   ->  0 errors
```

The 4 failures are in `tests/rapidata_client/audience/` and reproduce identically on `main` with these changes stashed — they are pre-existing and unrelated.

The repro above, with this branch: `tracer._enabled = False`, nothing exported. Without mocks imported: `tracer._enabled = True`, exporter live.

## Note on already-released versions

This only helps releases that ship it. Versions already in the wild (the bursts above include 3.16.7 and 3.23.1) will keep exporting, so a collector-side filter is worth having as well — opened separately against `infrastructure`.

🔗 Session: https://poseidon.rapidata.internal/chat/node-3d5691276069


RapidPoseidon:
Superseded by #893, which fixes the cause instead of adding a third process-detection heuristic: the tracer and OTLP log handler stay inactive until a `RapidataClient` is constructed, so components driven with mocked services never export. Suggest closing this one once #893 lands.

PR #896, opened 2026-09-22
feat(experiment): internal experiment creation and experiment jobs

manueltollis:
## Summary

Implements the python-sdk slice of RFC "Run experiments on direct SDK jobs" (`8e5e8608-9a8e-482c-8025-dd64c11407f2`), against the contract merged in `rapidata-backend#5378` (squash `19ce148`, prod tag `2026.09.22.0935-19ce148`).

- New private `rapi._experiment` surface (`ExperimentManager`), wired in `RapidataClient.__init__`:
  - `create(*, name, treatment_bps=5000, flags=None, user_enrollment=STICKY, description="") -> Experiment` — posts `scope="attached"`, then activates the created experiment by id.
  - `run_experiment(config: ExperimentConfig) -> RapidataJob` — raises on an empty `experiment_id`, else attaches it to the job via `audience._create_job`.
- New frozen `ExperimentConfig(job_definition, experiment_id, audience)` dataclass.
- `RapidataAudienceBase.assign_job` refactored into `_create_job(job_definition, experiment_id=None)`; `assign_job`'s public signature and behavior (warnings, printed job link) are unchanged, and it now delegates to `_create_job` with no experiment id.
- `RapidataJob` gains a read-only `experiment_id` property, populated from the job-creation response.
- OpenAPI client regeneration: merged the bot's already-open, CI-green `#892` (`chore: update OpenAPI client to 2026.09.22.0935-19ce148`) into `main` first and branched off that, rather than re-running the generator — its diff already carried `experimentId`, the `ExperimentApi`, `ExperimentScope`, etc. from the same prod deploy this PR needs.
- Added `experiment_api` to `CampaignService` (hand-written wiring, not generated) so `openapi_service.campaign.experiment_api` resolves.

### Not exported / undocumented, as specified
`ExperimentManager` and `ExperimentConfig` are not exported from the public `rapidata` package root and are used only via the private `self._experiment` attribute.

**Caveat:** mkdocs's `filters: ["!^_"]` hides underscore-prefixed *members* within a rendered page, but `scripts/gen_ref_pages.py` still generates a standalone reference page for any module file that isn't itself underscore-prefixed — so `ExperimentManager`/`Experiment`/`ExperimentConfig` do still surface on their module's generated docs page (confirmed via a local `mkdocs build`). This is a pre-existing gap, not a regression: `DemographicManager` (also private, via `self._demographic`) has the exact same doc-page leak today on `main`. Filing this here rather than silently reworking the shared doc-generation script, which would also touch every other private manager's docs.

## Test plan
- [x] New unit tests in `tests/rapidata_client/experiment/test_experiment_manager.py`: `create()` posts `scope=attached` then activates by id; `run_experiment()` raises on empty `experiment_id`; `run_experiment()` posts to `config.audience.id` with `experimentId` set and populates `job.experiment_id`.
- [x] New/updated unit tests in `tests/rapidata_client/audience/test_audience_base.py`: `assign_job` still sends no `experimentId`; `_create_job(..., experiment_id=...)` sends it and sets `job.experiment_id`.
- [x] `uv run python -m pytest` (full suite, matching CI's invocation): 284 passed.
- [x] `uv run black src/rapidata/rapidata_client` — files touched by this change are formatted (unrelated pre-existing formatting drift in other files was left untouched).
- [x] `uv run pyright src/rapidata/rapidata_client` — 0 errors.
- [x] `python3 -c "from rapidata import RapidataClient; print('OK')"` — OK.
- [x] `uv run --group docs mkdocs build` — succeeds; confirmed the doc-page leak described above.

Session: https://poseidon.rapidata.internal/chat/node-9b5f8b3b


src/rapidata/rapidata_client/job/rapidata_job.py
```diff
@@ -76,9 +77,15 @@ def __init__(
...
         self.__completed_at = None
         self.__estimated_cost: CostEstimate | None = None
+        self.__experiment_id = experiment_id
         self.job_details_page = f"https://app.{self._openapi_service.environment}/audiences/{self.audience_id}/job/{self.id}"
         logger.debug("RapidataJob initialized")
 
+    @property
+    def experiment_id(self) -> str | None:
+        """Internal: id of the experiment attached to this job's campaign, if any."""
+        return self.__experiment_id
```
LinoGiger: no, do not expose any of this

LinoGiger (changes requested):

src/rapidata/rapidata_client/job/rapidata_job.py
RapidPoseidon: Agreed, removed in e6a9ec6.

I took "any of this" to mean the whole experiment surface on `RapidataJob`, not just the property, so both are gone:

- The read-only `experiment_id` property is deleted.
- The `experiment_id` constructor kwarg is deleted too. It was still showing up in the rendered class signature on the generated reference page, since `merge_init_into_class` folds `__init__` into the class heading.

The id is now a plain private attribute, initialised to `None` in `__init__` and set by `audience._create_job` after construction. `RapidataJob`'s public signature is back to exactly what it was before this PR. Rebuilt the docs to confirm: the only remaining mention of "experiment" on that page is the private attribute assignment inside the collapsed source listing, which is how every private attribute in the SDK already renders.

Worth flagging for @manueltollis as RFC owner: the RFC's program design specifies `RapidataJob(..., experiment_id=response.experiment_id)` with a "new kwarg + read-only property", and its product section shows `>>> job.experiment_id` as the usage. This change deliberately diverges from that. Internal callers and the tests read `job._experiment_id` instead. Happy to reshape it if you wanted the value reachable some other way.

manueltollis: Yeah I don't really care about these things to be exposed to everyone, this is only for internal usage for us running the experiments

LinoGiger (approved):

